PR #1, opened 2016-08-04
Removes bearer variable from stack

elliemdaw:


acvp_transport.c
```diff
@@ -54,12 +54,14 @@ static long acvp_curl_http_get (ACVP_CTX *ctx, char *url, void *writefunc)
     long http_code = 0;
     CURL *hnd;
     struct curl_slist *slist;
-    char bearer[1048]; //TODO: remove this from the stack
 
     slist = NULL;
     if (ctx->jwt_token) {
-        snprintf(bearer, 1048, "Authorization: Bearer %s", ctx->jwt_token);
+    	int buf_size = strlen(ctx->jwt_token) + 22;
```
jfigus: We should probably use strnlen() instead of strlen(), then use a #define to declare a maximum JWT length that we'll allow.  Keep in mind the JWT is coming off the wire from the server.  It's best to never trust data coming off the wire.

elliemdaw: @jfigus I will update that :) is there a certain max length that would work well here?

jfigus: Let's just set it to 512 bytes for now.  The ACVP spec is no where near complete.  The content of the JWT may still be changing.

elliemdaw: Awesome, I updated that

jfigus:
Thx for the contribution.

PR #12, opened 2017-04-18
Adds HMAC capability

elliemdaw:


src/acvp.c
```diff
@@ -100,10 +105,18 @@ ACVP_ALG_HANDLER alg_tbl[ACVP_ALG_MAX] = {
...
-
     {ACVP_HASHDRBG,        &acvp_drbg_kat_handler,  ACVP_ALG_HASHDRBG},
     {ACVP_HMACDRBG,        &acvp_drbg_kat_handler,  ACVP_ALG_HMACDRBG},
     {ACVP_CTRDRBG,         &acvp_drbg_kat_handler,  ACVP_ALG_CTRDRBG},
+    {ACVP_HMAC_SHA1,       &acvp_hmac_kat_handler,  ACVP_ALG_HMAC_SHA1},
+    {ACVP_HMAC_SHA2_224,   &acvp_hmac_kat_handler,  ACVP_ALG_HMAC_SHA2_224},
+    {ACVP_HMAC_SHA2_256,   &acvp_hmac_kat_handler,  ACVP_ALG_HMAC_SHA2_256},
+    {ACVP_HMAC_SHA2_384,   &acvp_hmac_kat_handler,  ACVP_ALG_HMAC_SHA2_384},
+    {ACVP_HMAC_SHA2_512,   &acvp_hmac_kat_handler,  ACVP_ALG_HMAC_SHA2_512},
+    {ACVP_HMAC_SHA3_224,   &acvp_hmac_kat_handler,  ACVP_ALG_HMAC_SHA3_224},
```
bfussell: We discussed including SHA3 entries, but I may not have mentioned we may want to make sure they don't blow up in case someone tries to use them, ie if some ACVP user has SHA3 in their module and tries to register we want to return some error that says not supported.

src/acvp.c
```diff
@@ -1499,6 +1599,17 @@ static ACVP_RESULT acvp_build_register(ACVP_CTX *ctx, char **reg)
             case ACVP_CTRDRBG:
                 acvp_build_drbg_register_cap(cap_obj, cap_entry);
                 break;
+            case ACVP_HMAC_SHA1:
+            case ACVP_HMAC_SHA2_224:
+            case ACVP_HMAC_SHA2_256:
+            case ACVP_HMAC_SHA2_384:
+            case ACVP_HMAC_SHA2_512:
+            case ACVP_HMAC_SHA3_224:
```
bfussell: Same comment here - for HMAC-SHA3 we may want an unsupported algorithm return code here

src/acvp_lcl.h
```diff
@@ -65,6 +65,17 @@
 #define ACVP_ALG_HASHDRBG            "hashDRBG"
 #define ACVP_ALG_HMACDRBG            "hmacDRBG"
 #define ACVP_ALG_CTRDRBG             "ctrDRBG"
+#define ACVP_ALG_HMAC_SHA1           "HMAC-SHA1"
+#define ACVP_ALG_HMAC_SHA2_224       "HMAC-SHA2-224"
+#define ACVP_ALG_HMAC_SHA2_256       "HMAC-SHA2-256"
+#define ACVP_ALG_HMAC_SHA2_384       "HMAC-SHA2-384"
+#define ACVP_ALG_HMAC_SHA2_512       "HMAC-SHA2-512"
+#define ACVP_ALG_HMAC_SHA2_512_224   "HMAC-SHA2-512/224"
```
bfussell: Do we need 512/224 and 512/256 in the tables and case statements like SHA3 ?
And like SHA3 we return unsupported on registration until someone(not us) implements it ?

Makefile
```diff
@@ -3,7 +3,7 @@ CFLAGS+=-g -O0 -fPIC -Wall
 LDFLAGS+=
 INCDIRS+=-I. -Isrc
 
-SOURCES=src/acvp.c src/acvp_aes.c src/acvp_des.c src/acvp_hash.c src/acvp_transport.c src/acvp_util.c src/parson.c src/acvp_drbg.c
```
bfussell: Add the new file to Makefile.fom and Makefile.murl as well.

app/app_main.c
```diff
@@ -239,183 +242,197 @@ int main(int argc, char **argv)
...
-    rv = acvp_enable_sym_cipher_cap_parm(ctx, ACVP_TDES_CFB64, ACVP_SYM_CIPH_IVLEN, 192/3);
-    CHECK_ENABLE_CAP_RV(rv);
-    rv = acvp_enable_sym_cipher_cap_parm(ctx, ACVP_TDES_CFB64, ACVP_SYM_CIPH_PTLEN, 64 * 5);
-    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_enable_hmac_cap(ctx, ACVP_HMAC_SHA1, &app_hmac_handler);
+    rv = acvp_enable_hmac_cap(ctx, ACVP_HMAC_SHA2_224, &app_hmac_handler);
+    rv = acvp_enable_hmac_cap(ctx, ACVP_HMAC_SHA2_256, &app_hmac_handler);
+    rv = acvp_enable_hmac_cap(ctx, ACVP_HMAC_SHA2_384, &app_hmac_handler);
+    rv = acvp_enable_hmac_cap(ctx, ACVP_HMAC_SHA2_512, &app_hmac_handler);
+    rv = acvp_enable_hmac_cap(ctx, ACVP_HMAC_SHA3_224, &app_hmac_handler);
```
bfussell: Should we expect this to pass for SHA3 ? If so omit it for now and just register the SHA2 flavors we support.
ALL of them need a return code check, ie CHECK_ENABLE_CAP_RV(rv)

app/app_main.c
```diff
@@ -1289,6 +1306,65 @@ static ACVP_RESULT app_sha_handler(ACVP_TEST_CASE *test_case)
...
+      md = EVP_sha256();
+      break;
+    case ACVP_HMAC_SHA2_384:
+      md = EVP_sha384();
+      break;
+    case ACVP_HMAC_SHA2_512:
+      md = EVP_sha512();
+      break;
+    default:
+    	printf("Error: Unsupported hash algorithm requested by ACVP server\n");
```
bfussell: Looks like you have the unsupported error I was looking for in the other comment - so disregard that :)

PR #13, opened 2017-04-20
Adds ability to enable params for HMAC

elliemdaw:


src/acvp.c
```diff
@@ -449,6 +473,110 @@ ACVP_RESULT acvp_enable_hmac_cap(
     return (acvp_append_hmac_caps_entry(ctx, cap, cipher, crypto_handler));
 }
 
+/*
+ * The user should call this after invoking acvp_enable_sym_cipher_cap()
```
bfussell: enable_sym - is that a typo ?

src/acvp.c
```diff
@@ -1058,20 +1186,32 @@ static ACVP_RESULT acvp_build_hash_register_cap(JSON_Object *cap_obj, ACVP_CAPS_
...
+    json_object_set_value(cap_obj, "keyRange2", json_value_init_array());
+    temp_arr = json_object_get_array(cap_obj, "keyRange2");
+    json_array_append_number(temp_arr, cap_entry->cap.hmac_cap->key_range_2[0]);
+    json_array_append_number(temp_arr, cap_entry->cap.hmac_cap->key_range_2[1]);
+
+    json_object_set_string(cap_obj, "keyBlock", cap_entry->cap.hmac_cap->key_block ? "yes" : "no" );
+    json_object_set_string(cap_obj, "inEmpty", cap_entry->cap.hmac_cap->in_empty ? "yes" : "no" );
+
+    /*
+     * Set the supported AAD lengths (for AEAD ciphers)
```
bfussell: cut/paste leftover AES references

app/app_main.c
```diff
@@ -429,16 +429,102 @@ int main(int argc, char **argv)
     /*
      * Enable HMAC
      */
+     char value[] = "test_prereq_val";
```
bfussell: Since the valid values are "same" or a numerical value, you may want to just put "same" here so the server will accept it.

PR #16, opened 2017-04-27
Clean up registration for demo

bfussell:


app/app_main.c
```diff
@@ -221,7 +221,7 @@ int main(int argc, char **argv)
     /*
      * Setup the vendor attributes
      */
-    rv = acvp_set_vendor_info(ctx, "ACME company", "looneytunes.org", "Porky Pig", "pp@looneytunes.org");
+    rv = acvp_set_vendor_info(ctx, "Cisco Systems", "www.cisco.com", "Barry Fussell", "bfussell@cisco.com");
```
ashman-p: Was wondering if we should read in all the config parameters including vendor, server IP, port etc. Then there is no need to recompile for those kind of changes.

PR #18, opened 2017-05-11
Fixes memory resulting from json library method call

elliemdaw:
Sorry for the noise, my editor converts tabs to spaces... maybe there is an ignore whitespace option on github like there is on bitbucket?

The actual fix is around the call to json_serialize_to_string_pretty in each acvp_*.c file

### DIFF
--- a/src/acvp_drbg.c
+++ b/src/acvp_drbg.c
@@ -78,6 +78,7 @@ ACVP_RESULT acvp_drbg_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
 
     char            *der_func_str;
     char            *pred_resist_str;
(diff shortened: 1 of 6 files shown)


src/acvp_drbg.c
```diff
@@ -230,7 +231,7 @@ ACVP_RESULT acvp_drbg_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
             testval = json_array_get_value(tests, j);
             testobj = json_value_get_object(testval);
 
-            reg = json_serialize_to_string_pretty(testval);
```
bfussell: Should there be a free before using reg again here ?

PR #19, opened 2017-05-15
Hmac msglen fix

elliemdaw:
If msglen isn't included in the server response, the msglen will be set in the kat handler.

This PR also renames a few variables to be consistent with HMAC.

### DIFF
--- a/src/acvp_hmac.c
+++ b/src/acvp_hmac.c
@@ -212,7 +212,9 @@ ACVP_RESULT acvp_hmac_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
              * TODO: this does mallocs, we can probably do the mallocs once for
              *       the entire vector set to be more efficient
              */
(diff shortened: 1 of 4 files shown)


src/acvp_hmac.c
```diff
@@ -212,6 +212,9 @@ ACVP_RESULT acvp_hmac_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
              * TODO: this does mallocs, we can probably do the mallocs once for
              *       the entire vector set to be more efficient
              */
+            if (msglen == 0) {
```
ashman-p: i did not see where msglen was initialized.

elliemdaw: On line 189 it tries to grab it from the json response :)

ashman-p: OK, Thanks.

PR #22, opened 2017-06-09
Rsa keygen

elliemdaw:
This PR adds the RSA keygen functionality. The JSON advertisement for keygen registration matches the current NIST spec.

This PR does not include handling test vectors using the RSA API (rsa_handler in app_main is not functionally built out), but does include receiving and parsing the vectors. This functionality has not been thoroughly tested yet.


src/acvp.h
```diff
@@ -425,6 +476,63 @@ typedef struct acvp_cmac_tc_t {
...
+    unsigned char *prime_result; // "prime" or "composite"
+} ACVP_RSA_KEYGEN_TC;
+
+typedef struct acvp_rsa_tc_t {
+    ACVP_CIPHER cipher;
+    ACVP_RSA_MODE mode; // "keyGen"
+    int info_gen_by_server;
+    unsigned int rand_pq;
+    unsigned int mod;
+    ACVP_RSA_KEYGEN_TC *keygen_tc;
```
bfussell: I assume the plan is to add SigGen and SigVer TC structs here when they are coded ?

src/acvp.c
```diff
@@ -1196,6 +1195,402 @@ static ACVP_RESULT acvp_add_drbg_prereq_val(ACVP_DRBG_CAP_MODE *drbg_cap_mode,
...
+    ACVP_RSA_CAP_MODE_LIST *current_rsa_cap_list;
+
+    /*
+     * Validate input
+     */
+    if (!ctx) {
+        return ACVP_INVALID_ARG;
+    }
+
+    switch (cipher) {
```
bfussell: Are there more plans for this check or would a simple if (cipher != ACVP_RSA) suffice ?

src/acvp.c
```diff
@@ -1450,6 +1845,68 @@ ACVP_RESULT acvp_enable_drbg_cap(
...
 }
 
+ACVP_RESULT acvp_enable_rsa_prereq_cap(ACVP_CTX          *ctx,
+                                       ACVP_CIPHER       cipher,
+                                       ACVP_RSA_PRE_REQ pre_req,
+                                       char              *value)
+{
+    ACVP_CAPS_LIST          *cap_list;
+
+    if (!ctx) {
```
bfussell: May want to use ACVP_NO_CTX on these checks

src/acvp.c
```diff
@@ -1196,6 +1195,402 @@ static ACVP_RESULT acvp_add_drbg_prereq_val(ACVP_DRBG_CAP_MODE *drbg_cap_mode,
...
+                                          ACVP_RSA_MODE mode,
+                                          ACVP_CAPS_LIST **cap_list,
+                                          ACVP_RSA_CAP_MODE_LIST      **rsa_cap_mode_list) {
+
+    ACVP_RSA_CAP_MODE_LIST *current_rsa_cap_list;
+
+    /*
+     * Validate input
+     */
+    if (!ctx) {
```
bfussell: ACVP_NO_CTX here ?

src/acvp.c
```diff
@@ -1196,6 +1195,402 @@ static ACVP_RESULT acvp_add_drbg_prereq_val(ACVP_DRBG_CAP_MODE *drbg_cap_mode,
...
+        rsa_cap_mode_list->cap_mode_attrs.keygen->pub_exp = value;
+        break;
+    case ACVP_RAND_PQ:
+        rsa_cap_mode_list->cap_mode_attrs.keygen->rand_pq = value;
+        break;
+    case ACVP_RSA_INFO_GEN_BY_SERVER:
+        rsa_cap_mode_list->cap_mode_attrs.keygen->info_gen_by_server = value;
+        break;
+    default:
+        return ACVP_INVALID_ARG;
```
bfussell: I know we aren't real consistent on this, but an ACVP_LOG_ERR may help debug here.

src/acvp.h
```diff
@@ -251,6 +254,54 @@ typedef enum acvp_drbg_pre_req {
...
+#define MOD_PRIME_3072     3072
+#define MOD_PRIME_4096     4096
+
+#define PROB_PRIME_TEST_2       2
+#define PROB_PRIME_TEST_3       3
+
+#define PRIME_TEST_TBLC2_NAME "tblC2"
+#define PRIME_TEST_TBLC3_NAME "tblC3"
+
+#define RSA_RAND_PQ_B32        1
```
bfussell: Not a code review comment, but I wonder why we don't use B.3.2, B.3.3, etc. here so the JSON is not so cryptic, yes, intended :)

src/acvp.h
elliemdaw: That is the plan :)

src/acvp.h
elliemdaw: I defined them as ints here and wrote a mapping in acvp_util to get the string when needed. The reason was just to reduce duplicate code because that would have required an enable_parm function that took a char* as a parameter. Unless there is a better way?

bfussell: I was referring to the spec.  Section 2.4.1.3 we use {"1","2","3","4","5"} instead of being more explicit. 
Maybe that changed ?
Just ignore this for now, it has nothing to do with the code you wrote and more to do with the spec.

elliemdaw: Ohhhh I thought that changed/was changing with the most recent updates... I had it as integers but changed it haha let me see if I can find the diff

elliemdaw: They were updated to strings this morning :) line 660 https://github.com/usnistgov/ACVP/blob/master/artifacts/acvp_sub_rsa.txt

bfussell: Great, that makes more sense. Thanks for checking.

PR #24, opened 2017-06-12
Add KDF TLS

bfussell:


app/app_main.c
```diff
@@ -1678,6 +1699,129 @@ static ACVP_RESULT app_hmac_handler(ACVP_TEST_CASE *test_case)
...
+	break;
+    case ACVP_KDF135_TLS_CAP_SHA384:
+        evp_md1 = evp_md2 = EVP_sha384();
+	break;
+    case ACVP_KDF135_TLS_CAP_SHA512:
+        evp_md1 = evp_md2 = EVP_sha512();
+	break;
+    default:
+        printf("\nCrypto module error, Bad SHA type\n");
+        return ACVP_CRYPTO_MODULE_FAIL;
```
ashman-p: Should this error be something like unsupported algorithm?

app/app_main.c
```diff
@@ -1678,6 +1699,129 @@ static ACVP_RESULT app_hmac_handler(ACVP_TEST_CASE *test_case)
...
+            printf("\nCrypto module error, TLS kdf failure\n");
+            return ACVP_CRYPTO_MODULE_FAIL;
+        }
+        for (i = 0; i < olen1; i++) {
+            master_secret1[i] ^= master_secret2[i];
+        }
+    }
+
+
+    memset(key_block1, 0, 4096);
```
ashman-p: I guess eventually you may want to add a macro for the magic numbers?

bfussell: Good catch, I should clear that in the acvp code not the app code where I already have the size.

app/app_main.c
bfussell: Yep, I'll change that

PR #25, opened 2017-06-15
Feature rsa sigver

ashman-p:
Refactor RSA Key Gen to better accommodate changes for Sig Gen and Sig Ver.

### DIFF
--- a/src/acvp_rsa.c
+++ b/src/acvp_rsa.c
@@ -284,8 +284,8 @@ acvp_kat_rsa_keygen(int info_gen_by_server, int rand_pq, ACVP_CAPS_LIST* cap,
     return info_gen_by_server;
 }
 
-ACVP_RESULT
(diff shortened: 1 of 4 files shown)


src/acvp_rsa.c
```diff
@@ -236,6 +237,101 @@ static ACVP_RESULT acvp_rsa_release_tc(ACVP_RSA_TC *stc)
...
+        case RSA_RAND_PQ_B33:
+        default:
+            break;
+        }
+    }
+    return info_gen_by_server;
+}
+
+ACVP_RESULT
+acvp_process_keygen(int rand_pq, unsigned int mod, int info_gen_by_server,
```
elliemdaw: Could we call this 'prepare' or 'setup' or something instead of process since the actual processing happens once its passed to the crypto handler callback method? :)

ashman-p: I will use 'prep'. I was searching for a good name. Thanks.

PR #28, opened 2017-07-07
RSA Sig Verify Handler

ashman-p:
Code to implement RSA Sig Ver functionality.

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -1987,6 +1987,148 @@ static ACVP_RESULT app_rsa_handler(ACVP_TEST_CASE *test_case)
     return ACVP_SUCCESS;
 }
 
+/* RSA SigVer handler
+ * msg and sig are binary format
+ * exponent and modulus are hex values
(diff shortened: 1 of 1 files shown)


app/app_main.c
```diff
@@ -1987,6 +1987,148 @@ static ACVP_RESULT app_rsa_handler(ACVP_TEST_CASE *test_case)
...
+        break;
+    case ACVP_SHA512:
+        tc_md = (EVP_MD *)EVP_sha512();
+        break;
+    default:
+        printf( "ERROR: Unsupported hash.\n");
+        goto err;
+    }
+
+    ERR_load_crypto_strings();
```
bfussell: Is this something we need to do somewhere generically ?

ashman-p: Will move it up in main() somewhere.

ashman-p: The call is already in main(), so will remove this call.

PR #29, opened 2017-07-08
Common prereq

elliemdaw:
Collapses repeated prereq code into common structs and methods

DRBG prereqs are not included in this because the prereqs for DRBG are added according to each mode and therefore exist at a different level


src/acvp.c
```diff
@@ -2032,7 +1918,7 @@ ACVP_RESULT acvp_set_cacerts(ACVP_CTX *ctx, char *ca_file)
     /*
      * Enable peer verification when CA certs are provided.
      */
-    ctx->verify_peer = 1;
+    ctx->verify_peer = 0;
```
bfussell: Remove this change before you commit :)

elliemdaw: Ahh good catch. Fixed :)

PR #36, opened 2017-07-28
TDES: CFB1, CFB8, CFB64 and OFB

bfussell:
Updates to get Monte Carlo tests to work and some formatting.


app/app_main.c
```diff
@@ -1136,56 +1136,66 @@ static ACVP_RESULT app_des_handler(ACVP_TEST_CASE *test_case)
     if (tc->test_type == ACVP_SYM_TEST_TYPE_MCT) {
```
elliemdaw: Is there an easy way to combine the if/else here and just check if its the first round of MCT? It looks like other than that case, the API calls are the same.

PR #39, opened 2017-08-18
Kdf ssh branch

ashman-p:
Added client support for KDF SSH.


app/app_main.c
```diff
@@ -730,6 +731,23 @@ int main(int argc, char **argv)
    CHECK_ENABLE_CAP_RV(rv);
    rv = acvp_enable_prereq_cap(ctx, ACVP_KDF135_SNMP, ACVP_PREREQ_SHA, value);
    CHECK_ENABLE_CAP_RV(rv);
+
+   rv = acvp_enable_kdf135_ssh_cap(ctx, &app_kdf135_ssh_handler);
+   CHECK_ENABLE_CAP_RV(rv);
+   rv = acvp_enable_prereq_cap(ctx, ACVP_KDF135_SSH, ACVP_PREREQ_SHA, value);
```
bfussell: Wouldn't AES and TDES also be prerequisites ?

ashman-p: Make sense however, was not in the spec. in section 2.1. I will update the code.

src/acvp_kdf135_ssh.c
```diff
@@ -0,0 +1,386 @@
...
+                                    unsigned int hash_len,
+                                    const unsigned char *session_id,
+                                    unsigned int session_len)
+{
+    memset(stc, 0x0, sizeof(ACVP_KDF135_SSH_TC));
+
+    stc->shared_sec_k = calloc(1, ACVP_KDF135_SSH_MSG_MAX);
+    if (!stc->shared_sec_k) return ACVP_MALLOC_FAIL;
+    stc->hash_h = calloc(1, ACVP_KDF135_SSH_MSG_MAX);
+    if (!stc->hash_h) return ACVP_MALLOC_FAIL;
```
nayrpg: Hey Norm, I see where shared_sec_k and hash_h are given space but I don't see where the information for them is put in the buffer you created.

nayrpg: I would think you would need to do a strcpy or the like to move the information from the passed in hash_h and shared_sec_k to the test case buffer locations

ashman-p: Good catch. Fixed.

src/acvp_kdf135_ssh.c
```diff
@@ -0,0 +1,389 @@
...
+    if (!stc->sc_e_key) return ACVP_MALLOC_FAIL;
+
+    stc->cs_i_key = calloc(1, ACVP_KDF135_SSH_MSG_MAX);
+    if (!stc->cs_i_key) return ACVP_MALLOC_FAIL;
+    stc->sc_i_key = calloc(1, ACVP_KDF135_SSH_MSG_MAX);
+    if (!stc->sc_i_key) return ACVP_MALLOC_FAIL;
+
+    memcpy(stc->shared_sec_k, shared_sec_k, key_len);
+    memcpy(stc->hash_h, hash_h, hash_len);
+
```
nayrpg: You need one for the session_id as well I assume. Also is it possible for the incoming JSON key_len property to be longer than the ACVP_KDF135_SSH_MSG_MAX. If it is we should probably add some verification of the length to be sure that we don't write passed the buffer's size.

PR #40, opened 2017-08-25
Adding RSA SigGen, SigVer capability registration, and Multiple SigType registrations

nayrpg:
This pull adds in Melissa Novitsky's code which provides RSA SigGen capability registration, and test vector support.
It also adds Ryan Granger's code which adds RSA SigVer capability registration and registration of multiple RSA mode capabilities (e.g. RSA SigGen and SigVer enabled simultaneously). This pull request is a second attempt. See the #38 for previous failed attempt


src/acvp_rsa.c
```diff
@@ -332,6 +438,74 @@ acvp_prep_keygen(int rand_pq, unsigned int mod, int info_gen_by_server,
...
+         * Setup the test case data that will be passed down to
+         * the crypto module.
+         * TODO: this does mallocs,kat we can probably do the mallocs once for
+         *       the entire vector set to be more efficient
+         */
+        rv = acvp_rsa_init_sig_tc(ctx, sigtc, tc_id, alg_id, /* note: mode is set in kat_handler */
+                sig_type, modulo, hash_alg, msg, salt_len); /* siggen attrs */
+        break;
+    }
+//    case ACVP_RSA_MODE_SIGVER:
```
ashman-p: Should this case be removed or completed?

src/acvp_util.c
```diff
@@ -462,6 +464,76 @@ ACVP_RSA_CAP_MODE_LIST* acvp_locate_rsa_mode_entry(ACVP_CAPS_LIST *cap,
     return NULL;
 }
 
+ACVP_RSA_CAP_MODE_LIST* acvp_locate_rsa_sig_type_entry(ACVP_CAPS_LIST *cap,
```
ashman-p: Just wondering if there should be a distinction between the return cases? Seems like the 'error' case and the 'not found' cases could be treated the same way. So sig_type_str error or others are treated as 'not found' which could cause an attempt to allocate a linked list entry.

Maybe we could define and create a dummy 'empty struct' when the object is not found and use 'null' for the error case?

nayrpg: Good catch it would make a new mode list with the bogus mode type. I am not sure which way we want to do things but I was planning on having that function say it either found what you were looking for or it didn't. I think that we should pull the mode string check outside of the search function and error there and keep that function simply as a search.

src/acvp_rsa.c
nayrpg: removed until the next pull request when I work on SigVer test case handling?

app/app_main.c
```diff
@@ -2338,6 +2470,143 @@ static ACVP_RESULT app_rsa_handler(ACVP_TEST_CASE *test_case)
...
+    }
+
+    /*
+     * Make an RSA object and set a new BN exponent to use to generate a key
+     */
+    if (rsa) {
+        FIPS_rsa_free(rsa);
+    }
+
+    rsa = FIPS_rsa_new();
```
bfussell: Have we determined if this test is not available at runtime ? In general if you can perform it at runtime you don't use the FIPS calls, you would use the standard openssl call such as RSA_new() in this case.

src/acvp.c
```diff
@@ -1964,7 +2599,7 @@ ACVP_RESULT acvp_set_cacerts(ACVP_CTX *ctx, char *ca_file)
     /*
      * Enable peer verification when CA certs are provided.
      */
-    ctx->verify_peer = 1;
+    ctx->verify_peer = 0;
```
bfussell: flip this back before merging :)

nayrpg:
I cannot merge the branch. I think someone else will have to do it or you guys could add me as a collaborator.

PR #42, opened 2017-09-13
Initial TDES-OFB and Keywrap commit

bfussell:
Still needs more testing with NIST server, but functionally working.

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -73,6 +73,7 @@ int dsa_builtin_paramgen(DSA *ret, size_t bits, size_t qbits,
 static ACVP_RESULT app_aes_handler_aead(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_aes_keywrap_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_aes_handler(ACVP_TEST_CASE *test_case);
+static ACVP_RESULT app_des_keywrap_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_des_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_sha_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_hmac_handler(ACVP_TEST_CASE *test_case);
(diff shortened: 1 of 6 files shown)


app/app_main.c
```diff
@@ -1562,47 +1563,98 @@ static ACVP_RESULT app_aes_keywrap_handler(ACVP_TEST_CASE *test_case)
...
+        if (c_len <= 0) {
+            printf("Error: key wrap operation failed (%d)\n", c_len);
+            return ACVP_CRYPTO_MODULE_FAIL;
+        } else {
+            tc->ct_len = c_len;
+        }
+    } else if (tc->direction == ACVP_DIR_DECRYPT) {
+        EVP_CIPHER_CTX_set_flags(&cipher_ctx, EVP_CIPHER_CTX_FLAG_WRAP_ALLOW);
+        EVP_CipherInit_ex(&cipher_ctx, cipher, NULL, tc->key, NULL, 0);
+        c_len = EVP_Cipher(&cipher_ctx, tc->pt, tc->ct, tc->ct_len);
```
elliemdaw: Is the behavior of EVP_Cipher and EVP_Cipher_Update different in these handlers?

src/acvp_des.c
```diff
@@ -690,15 +692,18 @@ ACVP_RESULT acvp_des_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
 
                 /* Process the current DES encrypt test vector... */
                 rv = (cap->crypto_handler)(&tc);
+
                 if (rv != ACVP_SUCCESS) {
-                    ACVP_LOG_ERR("crypto module failed the operation");
-                    return ACVP_CRYPTO_MODULE_FAIL;
+                    if (rv != ACVP_CRYPTO_WRAP_FAIL) {
```
elliemdaw: We don't fail on wrap failure?

bfussell: It is an expected failure.  The server will send some decrypts that don't work and we need to return
"decryptFail" : true rather than the plaintext.

app/app_main.c
bfussell: Good catch, while playing around with this trying to make it work I tried CIPHER_Update.  Still didn't work with the server.  So I know this needs work, in fact this test may be something we don't support - I have a question out to Harold on whether NIST uses the same TDES-KW that openssl supports.

PR #43, opened 2017-09-22
Interop with NIST server for HMAC

bfussell:
Some bit/byte changes and pass the maclen to the app in case some other
app may need it.


src/acvp.c
```diff
@@ -4559,7 +4559,7 @@ static ACVP_RESULT acvp_get_result_vsid(ACVP_CTX *ctx, int vs_id)
     JSON_Value *val;
     JSON_Object *obj = NULL;
     char *json_buf;
-    int retry_count = 2700; /* 15 minutes*/
+    int retry_count = 900; /* 15 minutes*/
```
rygrange: The only change I would make is to change the retry count comment (I assume it should now say 5 minutes). Other than that looks good

bfussell: 900 is 15 minutes. The 2700 had incorrectly been committed(probably by me) when I extended the time to help NIST debug TDES-OFB.

PR #45, opened 2017-12-14
Fix TDES-CFB1 for new testcase format

bfussell:


src/acvp_des.c
```diff
@@ -650,19 +650,23 @@ ACVP_RESULT acvp_des_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
                 if (!pt) return (ACVP_MALFORMED_JSON);
 
                 if (alg_id == ACVP_TDES_CFB1) {
-                    ptlen = strlen((char *)pt)*(8);
+                    ptlen = (unsigned int)json_object_get_number(testobj, "ptLen");
+                    if (!ptlen) 
+                        ptlen = strlen((char *)pt)*(8/2);
```
rygrange: Should the ptlen default to `strlen((char *)pt)*(8);` as it did before when the alg_id is ACVP_TDES_CFB1?

bfussell: The original code was wrong.  They changed the way CFB1 works so that ptLen/ctLen is optional. By that they mean it will carry the bit length UNLESS they want us to use the whole byte(s).  So if the number of bits is 6, ptLen : 6. If the number of bits is 8, then they will not send ptLen.  For 8 bits pt will look something like pt : 55.  Thats a strlen of 2 * (8/2) = 8 bits.

PR #46, opened 2018-01-30
Addresses valgrind and coverity issues

elliemdaw:
- Addresses memory leaks from libacvp
- Fixes static analysis errors
- Fixes build warnings

### DIFF
--- a/src/acvp_des.c
+++ b/src/acvp_des.c
@@ -385,7 +385,7 @@ static ACVP_RESULT acvp_des_mct_tc(ACVP_CTX *ctx, ACVP_CAPS_LIST *cap,
     }
 
 
-    for (i = 0; i < 400; ++i) {
+    for (i = 0; i < ACVP_DES_MCT_OUTER; ++i) {
 
         /*
          * Create a new test case in the response
@@ -403,7 +403,7 @@ static ACVP_RESULT acvp_des_mct_tc(ACVP_CTX *ctx, ACVP_CAPS_LIST *cap,
             return rv;
         }
 
-        for (j = 0; j < 10000; ++j) {
+        for (j = 0; j < ACVP_DES_MCT_INNER; ++j) {
 
             if (j == 0) {
                 memcpy(old_iv, stc->iv, stc->iv_len);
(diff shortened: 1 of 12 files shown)


src/acvp_des.c
```diff
@@ -630,6 +654,10 @@ ACVP_RESULT acvp_des_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
             key3 = (unsigned char *)json_object_get_string(testobj, "key3");
 
             /* TODO: remove this its there to work with our server */
+            if (key) {
```
bfussell: nit: shouldn't have to check for NULL before calling free

src/acvp_des.c
```diff
@@ -647,7 +675,13 @@ ACVP_RESULT acvp_des_kat_handler(ACVP_CTX *ctx, JSON_Object *obj)
             if (dir == ACVP_DIR_ENCRYPT) {
                 pt = (unsigned char *)json_object_get_string(testobj, "pt");
                 iv = (unsigned char *)json_object_get_string(testobj, "iv");
-                if (!pt) return (ACVP_MALFORMED_JSON);
+                if (!pt) {
+                    if (key) {
```
bfussell: same here

src/acvp_hash.c
```diff
@@ -105,6 +105,8 @@ static ACVP_RESULT acvp_hash_mct_tc(ACVP_CTX *ctx, ACVP_CAPS_LIST *cap,
     ACVP_RESULT rv;
     JSON_Value          *r_tval = NULL; /* Response testval */
     JSON_Object         *r_tobj = NULL; /* Response testobj */
+    int num_mct_outer = 100;
+    int num_mct_inner = 1000;
```
bfussell: should we create a #define for these ? ACVP_HASH_MCT_INNER, OUTER ?

elliemdaw:
@bfussell yes, spacing is yet to come. The one spot where some got cleaned up is because I couldn't figure out the memory allocation / free logic with how it was displaying in the editor, so I went ahead and cleaned that up :)

PR #47, opened 2018-01-31
Adds doxygen comments for remaining API methods

elliemdaw:
Adds doxygen documentation for remaining public API methods

### DIFF
--- a/src/acvp.h
+++ b/src/acvp.h
@@ -1014,21 +1014,21 @@ ACVP_RESULT acvp_enable_drbg_cap_parm(
                              );
 
 /*! @brief acvp_enable_drbg_prereq_cap() allows an application to specify
- *      a prerequisite algorithm for a given DRBG during a test session
- *      with the ACVP server.
- *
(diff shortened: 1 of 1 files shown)


src/acvp.h
```diff
@@ -1032,13 +1163,53 @@ ACVP_RESULT acvp_enable_rsa_cap_parm(
...
+    @return ACVP_RESULT
+ */
 ACVP_RESULT acvp_enable_rsa_bignum_parm (ACVP_CTX *ctx,
                              ACVP_CIPHER cipher,
                              ACVP_RSA_MODE mode,
                              ACVP_RSA_PARM param,
                              BIGNUM *value
                            );
 
+/*! @brief acvp_enable_rsa_primes_parm() allows an application to specify
```
bfussell: Do the function headers with leading '*' turn out the same as those without.
Probably want to be consistent if it impacts the end document.

elliemdaw: Removed the extras :)

PR #50, opened 2018-02-08
Build and test

elliemdaw:
- Adds Criterion UT framework and a few starter UTs
- Fixes issues from starter UTs


elliemdaw:
@bfussell Ahh good point, I will check with Paul. I went with Criterion because it is an active project whereas Unit and other frameworks haven't had releases in years. We should get that sorted out first though :)

PR #52, opened 2018-02-27
Updates for CMAC interop

elliemdaw:
- Adds proper support for CMAC-TDES
- Adjusts byte/string mismatches

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -733,14 +733,11 @@ static void enable_cmac (ACVP_CTX *ctx) {
     /*
      * Enable CMAC
      */
-#if 0 // CMAC-AES not quite ready...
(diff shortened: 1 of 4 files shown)


app/app_main.c
```diff
@@ -2095,12 +2091,23 @@ static ACVP_RESULT app_hmac_handler(ACVP_TEST_CASE *test_case)
     return ACVP_SUCCESS;
 }
 
+void printBytes(unsigned char *buf, size_t len) {
```
bfussell: Left over debug ?

elliemdaw: Oh whoops, thought I left that out :) I'll fix that!

elliemdaw: Fixed!

PR #56, opened 2018-03-14
RSA Interop

elliemdaw:
- Various changes to registration object
- Reorganizes RSA modes to be their own ciphers, similar to how the KDFs are implemented
- Adds files for sig and keygen kat handlers
- Updates app code to handle processing of the vectors (only for the modes we register for)


rjrelyea:
Hi all,

So I've been trying to update to the PKCS #11 driver from the old libacvp to the new one and I've ran into the issue that the new libacvp now has a dependency on the openssl BigNum library, which means libacvp must link with openssl even though your library under test isn't openssl. I think it should have some library independent way of passing the RSA components, either as straight hex (as it comes from the server) or as a simple byte array. All the libraries know how turn byte arrays into their own individual BigNums. Would you be willing to entertain a patch that does this?

Thanks,

bob

elliemdaw:
Hey Bob!

I actually just added this to my todo list earlier today... we are on the same page! I was thinking the easiest way to do this would be to leave the values as character arrays since that is what they are when we parse the JSON, then the crypto handler can do what it needs to do. Does that seem reasonable?

Thanks!
Ellie

rjrelyea:
Cool. Yes leaving them as hex char arrays seems reasonable. The avoids the double conversion hex->byte array->native BN.

app/app_main.c
```diff
@@ -2694,100 +2710,56 @@ static ACVP_RESULT app_kdf135_ssh_handler(ACVP_TEST_CASE *test_case)
...
-                                      seed, seed_len,
-                                      bitlen1, bitlen2, bitlen3, bitlen4,
-                                      exponent, keylen, NULL) != 1) {
-            return ACVP_CRYPTO_MODULE_FAIL;
-        }
-        tc->keygen_tc->p = rsa->p;
-        tc->keygen_tc->q = rsa->q;
-        tc->keygen_tc->n = rsa->n;
-        tc->keygen_tc->d = rsa->d;
+    RSA_free(rsa);
```
bfussell: Probably need to be consistent and use the FOM calls for free since we use FOM call FIPS_rsa_new()

src/acvp.c
```diff
@@ -4387,24 +4356,30 @@ static ACVP_RESULT acvp_process_vsid (ACVP_CTX *ctx, int vs_id) {
...
-
+    
     if (!alg) {
         ACVP_LOG_ERR("JSON parse error: ACV algorithm not found");
         return ACVP_JSON_ERR;
     }
 
     ACVP_LOG_STATUS("vsId: %d", vs_id);
     ACVP_LOG_STATUS("ACV Operation: %s", alg);
+    // TODO: make sure this is included where relevant only
```
bfussell: Is this for later or something you've already done ?

src/acvp_rsa.c
```diff
@@ -1,830 +0,0 @@
-/** @file */
```
bfussell: I assume this file was deleted ?

app/app_main.c
```diff
@@ -2723,7 +2719,10 @@ static ACVP_RESULT app_rsa_keygen_handler(ACVP_TEST_CASE *test_case)
     ACVP_RSA_KEYGEN_TC    *tc;
     ACVP_RESULT rv = ACVP_SUCCESS;
     RSA       *rsa;
-    
+    BIGNUM *e = BN_new();
```
bfussell: What determines whether you use FIPS_bn_new() vs BN_new() ?

PR #62, opened 2018-04-10
Add ecdsa

elliemdaw:
- This hasn't been fully tested with the NIST server since I haven't been able to successfully retrieve a vector set
- Adds the library side for ECDSA
- Has a stub for the ECDSA handler in the app code

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -78,6 +78,7 @@ static void enable_hmac(ACVP_CTX *ctx);
 static void enable_kdf(ACVP_CTX *ctx);
 static void enable_dsa(ACVP_CTX *ctx);
 static void enable_rsa(ACVP_CTX *ctx);
+static void enable_ecdsa(ACVP_CTX *ctx);
 static void enable_drbg(ACVP_CTX *ctx);
 
 static ACVP_RESULT app_aes_handler_aead(ACVP_TEST_CASE *test_case);
@@ -99,6 +100,7 @@ static ACVP_RESULT app_kdf135_ssh_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_drbg_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_rsa_keygen_handler(ACVP_TEST_CASE *test_case);
 static ACVP_RESULT app_rsa_sig_handler(ACVP_TEST_CASE *test_case);
+static ACVP_RESULT app_ecdsa_handler(ACVP_TEST_CASE *test_case);
 #endif
 
 #define DEFAULT_SERVER "127.0.0.1"
@@ -222,6 +224,7 @@ int main(int argc, char **argv)
      */
     int rsa = 0;
     int drbg = 0;
+    int ecdsa = 1;
 
     if (argc > 3) {
         print_usage();
@@ -377,6 +380,10 @@ int main(int argc, char **argv)
 #ifdef ACVP_NO_RUNTIME
     if (rsa) {
         enable_rsa(ctx);
+    }
+    
+    if (ecdsa) {
+        enable_ecdsa(ctx);
     }
 
     if (drbg) {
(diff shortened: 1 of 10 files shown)


app/app_main.c
```diff
@@ -1194,9 +1201,76 @@ static void enable_rsa (ACVP_CTX *ctx) {
...
+    rv = acvp_enable_ecdsa_cap_parm(ctx, ACVP_ECDSA_SIGVER, ACVP_CURVE, "k-163");
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_enable_ecdsa_cap_parm(ctx, ACVP_ECDSA_SIGVER, ACVP_CURVE, "b-163");
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_enable_ecdsa_cap_parm(ctx, ACVP_ECDSA_SIGVER, ACVP_HASH_ALG, "SHA2-224");
+    CHECK_ENABLE_CAP_RV(rv);
 #endif
+}
 
-#ifdef ACVP_NO_RUNTIME
```
bfussell: Is DRBG allowed at runtime now ? Or maybe blocked somewhere else ?

src/acvp_ecdsa.c
```diff
@@ -0,0 +1,324 @@
...
+        strncpy((char *)stc->s, s, strnlen(s, 128));
+        strncpy((char *)stc->r, r, strnlen(r, 128));
+    }
+    if (cipher == ACVP_ECDSA_SIGVER || cipher == ACVP_ECDSA_SIGGEN) {
+        stc->message = calloc(ACVP_RSA_EXP_LEN_MAX, sizeof(char));
+        if (!stc->message) { return ACVP_MALLOC_FAIL; }
+        strncpy((char *)stc->message, message, strnlen(message, ACVP_RSA_MSGLEN_MAX));
+    }
+    stc->d = calloc(ACVP_RSA_EXP_LEN_MAX, sizeof(char));
+    if (!stc->d) { return ACVP_MALLOC_FAIL; }
```
bfussell: Do we need to spit out an error message on the failure so we know what type it is or does the caller do that ?

src/acvp_ecdsa.c
```diff
@@ -0,0 +1,324 @@
...
+             */
+            r_tval = json_value_init_object();
+            r_tobj = json_value_get_object(r_tval);
+            
+            json_object_set_number(r_tobj, "tcId", tc_id);
+            
+            rv = acvp_ecdsa_init_tc(ctx, alg_id, &stc, tc_id, curve, secret_gen_mode, qx, qy, message, r, s);
+            
+            /* Process the current test vector... */
+            if (rv == ACVP_SUCCESS) {
```
bfussell: On failure should we exit out here ?

src/acvp_ecdsa.c
```diff
@@ -0,0 +1,324 @@
...
+    json_result = json_serialize_to_string_pretty(ctx->kat_resp);
+    if (ctx->debug == ACVP_LOG_LVL_VERBOSE) {
+        printf("\n\n%s\n\n", json_result);
+    } else {
+        ACVP_LOG_INFO("\n\n%s\n\n", json_result);
+    }
+    json_free_serialized_string(json_result);
+    
+    return rv;
+}
```
bfussell: no newline at end of file

src/acvp_ecdsa.c
```diff
@@ -0,0 +1,324 @@
...
+            }
+            
+            /*
+             * Release all the memory associated with the test case
+             */
+            key_err:
+            acvp_ecdsa_release_tc(&stc);
+            
+            /* Append the test response value to array */
+            json_array_append_value(r_tarr, r_tval);
```
bfussell: Should we try and append on a failure ?

app/app_main.c
elliemdaw: Oh that looks confusing in the diff... I made them all a part of the same #ifdef so DRBG, ECDSA, and RSA are all in the single block.

src/acvp_ecdsa.c
elliemdaw: I will add more verbose logs here :)

elliemdaw:
@bfussell I pushed up changes from these comments... the logging in acvp_ecdsa to initialize a test case is generic unless the calloc fails for the sig specific buffer

PR #66, opened 2018-04-24
Reorganizes alg/mode strings and alg_tbl

elliemdaw:


src/acvp.c
```diff
@@ -4796,36 +4795,27 @@ static ACVP_RESULT acvp_dispatch_vector_set (ACVP_CTX *ctx, JSON_Object *obj) {
...
-        if (!strncmp(alg_tbl_index, alg_tbl[i].name, strlen(alg_tbl[i].name))) {
-            rv = (alg_tbl[i].handler)(ctx, obj);
-            free((void *)alg_tbl_index);
-            return rv;
-        } else if (mode && !strncmp(mode, alg_tbl[i].name, strlen(alg_tbl[i].name))) {
-            rv = (alg_tbl[i].handler)(ctx, obj);
-            free((void *)alg_tbl_index);
-            return rv;
+        if (!strncmp(alg, alg_tbl[i].name, strlen(alg_tbl[i].name))) {
+            if (alg_tbl[i].mode != NULL) {
```
bfussell: May want to check to make sure mode is not null here before trying to do the strncm
That would cover the case where the server sent bad data, ie "RSA" but failed to send the mode such as "keyGen".

elliemdaw: Fixed :)

PR #70, opened 2018-04-27
Kdf135 srtp

elliemdaw:
Adds library support for KDF135 SRTP

Interop testing to come...

### DIFF
--- a/src/acvp_kdf135_srtp.c
+++ b/src/acvp_kdf135_srtp.c
@@ -32,47 +32,233 @@
 #include "parson.h"
 
 /*
- * Forward prototypes for local functions
- */
-static ACVP_RESULT acvp_kdf135_srtp_output_tc (ACVP_CTX *ctx, ACVP_KDF135_SRTP_TC *stc, JSON_Object *tc_rsp);
-
-static ACVP_RESULT acvp_kdf135_srtp_init_tc (ACVP_CTX *ctx,
-                                             ACVP_KDF135_SRTP_TC *stc,
-                                             unsigned int tc_id,
-                                             ACVP_CIPHER alg_id,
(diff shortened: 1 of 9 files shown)


src/acvp_kdf135_srtp.c
```diff
@@ -0,0 +1,264 @@
...
+    
+    if (!kdr || !master_key || !master_salt || !index || !srtcp_index) {
+        ACVP_LOG_ERR("Missing parameters - initalize KDF SRTP test case");
+        return ACVP_INVALID_ARG;
+    }
+    
+    stc->tc_id = tc_id;
+    stc->aes_keylen = aes_keylen;
+    
+    stc->kdr = calloc(8, sizeof(char));
```
bfussell: Do we want to make these hard-coded values #defines ?

src/acvp_lcl.h
```diff
@@ -58,7 +58,7 @@
 } while (0)
 #endif
 
-#define ACVP_ALG_MAX 61  /* Used by alg_tbl[] */
+#define ACVP_ALG_MAX ACVP_CIPHER_END  /* Used by alg_tbl[] */
```
bfussell: Thank you ! I've been meaning to do something with that and keep forgetting.

src/acvp_lcl.h
```diff
@@ -276,6 +277,12 @@ typedef struct acvp_kdf135_ssh_capability {
     int sha;
 } ACVP_KDF135_SSH_CAP;
 
+typedef struct acvp_kdf135_srtp_capability {
+    int supports_zero_kdr;
+    int kdr_exp[24];
```
bfussell: #define make sense here ?

PR #78, opened 2018-05-24
Kdf135 tpm

elliemdaw:
- Library support to add kdf tpm to registration
- This does not include interop
**TPM (and a couple of other TPMs) spec's aren't up to date, so I haven't fixed JSON naming, etc in the registration since the consistency is still to be addressed :)

### DIFF
--- a/src/acvp.h
+++ b/src/acvp.h
@@ -149,6 +149,7 @@ typedef enum acvp_cipher {
     ACVP_KDF135_SRTP,
     ACVP_KDF135_IKEV2,
     ACVP_KDF135_IKEV1,
+    ACVP_KDF135_TPM,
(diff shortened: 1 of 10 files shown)


src/acvp.h
```diff
@@ -167,7 +168,7 @@ typedef enum acvp_prereq_mode_t {
 } ACVP_PREREQ_ALG;
 
 #define ACVP_KDF135_SNMP_ENGID_MAX 32
-#define ACVP_KDF135_SNMP_SKEY_MAX 32
```
bfussell: I would like to keep these separate just in case the strength of one iKDF s changed and the other is not.
Otherwise everything else looks good !

PR #83, opened 2018-06-06
Adds updates from kdf135 tls and srtp interop testing

elliemdaw:


app/app_main.c
```diff
@@ -952,7 +952,7 @@ static void enable_kdf (ACVP_CTX *ctx) {
     /*
      * Enable KDF-135
      */
-    rv = acvp_enable_kdf135_tls_cap(ctx, ACVP_KDF135_TLS, &app_kdf135_tls_handler);
```
bfussell: Are we removing the KDF code from the app in github at some point ?

elliemdaw: Ahh so I hadn't added any of the new kdf code, but I'll take out the ones that are in there from earlier. Give me a minute :)

elliemdaw: Actually I think it may be good to leave what we have for reference. I had already taken out our API calls.

bfussell: OK, got it. registration is good, handler implementation bad.

PR #84, opened 2018-06-12
KAS-ECC for CDH-Component

bfussell:


src/acvp.h
```diff
@@ -1380,6 +1451,77 @@ ACVP_RESULT acvp_enable_dsa_cap_parm (ACVP_CTX *ctx,
...
+   @param cipher ACVP_CIPHER enum value identifying the crypto capability.
+   @param mode ACVP_KAS_ECC_MODE enum value specifying mode. An example would be
+        ACVP_KAS_ECC_MODE_PARTIAL
+   @param pre_req ACVP_PREREQ_ALG enum that the specified cipher/mode
+        depends on
+   @param value "same" or number
+
+   @return ACVP_RESULT
+ */
+ACVP_RESULT acvp_enable_kas_ecc_prereq_cap (
```
elliemdaw: Does this one have to separate from the general enable_prereq API?

Makefile.murl
```diff
@@ -25,6 +25,7 @@ SOURCES= src/acvp.c \
          src/acvp_kdf135_x963.c \
          src/acvp_kdf135_tpm.c \
          src/acvp_kdf108.c \
+         src/acvp_kas_ecc.c \
```
elliemdaw: I think Makefile.win may have gotten missed? :)

app/app_main.c
```diff
@@ -3399,6 +3449,218 @@ static ACVP_RESULT app_dsa_handler(ACVP_TEST_CASE *test_case)
     return ACVP_SUCCESS;
 }
 
+static EC_POINT *make_peer(EC_GROUP *group, BIGNUM *x, BIGNUM *y)
```
elliemdaw: Do we want to leave stubs here or keep these filled in?

Makefile.murl
elliemdaw: Also not sure if I am just missing it, but is acvp_kas_ecc included? It isn't showing up on my diff

app/app_main.c
```diff
@@ -3399,6 +3449,218 @@ static ACVP_RESULT app_dsa_handler(ACVP_TEST_CASE *test_case)
...
+
+
+    if (add_e) {
+        tc->pix = BN_bn2hex(tx);
+        tc->piy = BN_bn2hex(ty);
+        tc->pixlen = strlen(tc->pix);
+        tc->piylen = strlen(tc->piy);
+        if (d) {
+            tc->d = BN_bn2hex(d);
+            tc->dlen = strlen(tc->d);
```
fliphil: Strnlen instead of strlen? (general comment)

app/app_main.c
```diff
@@ -3399,6 +3449,218 @@ static ACVP_RESULT app_dsa_handler(ACVP_TEST_CASE *test_case)
...
+    if (exout)
+        d = EC_KEY_get0_private_key(key);
+    meth = EC_GROUP_method_of(grp);
+    if (EC_METHOD_get_field_type(meth) == NID_X9_62_prime_field)
+        rv = EC_POINT_get_affine_coordinates_GFp(grp, pt, tx, ty, ctx);
+    else
+#ifdef OPENSSL_NO_EC2M
+    {
+        printf("ERROR: GF2m not supported\n");
+        return 0;
```
fliphil: Need to free ctx before return

app/app_main.c
```diff
@@ -3399,6 +3449,218 @@ static ACVP_RESULT app_dsa_handler(ACVP_TEST_CASE *test_case)
...
+        return ACVP_CRYPTO_MODULE_FAIL;
+    }
+    Z = OPENSSL_malloc(Zlen);
+    if (!Z) {
+        printf("Malloc failure\n");
+        return ACVP_CRYPTO_MODULE_FAIL;
+    }
+    if (!ECDH_compute_key(Z, Zlen, peerkey, ec, 0)) {
+        printf("ECDH_compute_key failure\n");
+        return ACVP_CRYPTO_MODULE_FAIL;
```
fliphil: Need to free all of the allocated stuff before error returns (general comment)

app/app_main.c
```diff
@@ -3399,6 +3449,218 @@ static ACVP_RESULT app_dsa_handler(ACVP_TEST_CASE *test_case)
...
+    int nid = 0, exout = 0;
+    EC_KEY *ec = NULL;
+    EC_POINT *peerkey = NULL;
+    unsigned char *Z;
+    int Zlen;
+    BIGNUM *cx = NULL, *cy = NULL;
+
+    tc = test_case->tc.kas_ecc;
+
+    if (!strcmp(tc->curve, "b-233"))
```
fliphil: Thoughts on moving the strcmp into acvp_kas_ecc_kat_handler? Then make tc->curve an enum value. That way client app doesn't need to worry if ACVP spec JSON strings change.

src/acvp.c
```diff
@@ -349,7 +353,63 @@ static void acvp_cap_free_rsa_sig_list (ACVP_CAPS_LIST *cap_list) {
...
+                    do {
+                        next_curve = current_curve->next;
+                        free(current_curve);
+                        current_curve = next_curve;
+                    } while (current_curve);
+                }
+            }
+        }
+    }
+    free(cap_list->cap.kas_ecc_cap->kas_ecc_mode);
```
fliphil: Set to NULL after?

src/acvp.c
```diff
@@ -455,6 +515,10 @@ ACVP_RESULT acvp_free_test_session (ACVP_CTX *ctx) {
                     acvp_cap_free_dsa_attrs(cap_entry);
                     free(cap_entry->cap.dsa_cap);
                     break;
+                case ACVP_KAS_ECC_TYPE:
+                    acvp_cap_free_kas_ecc_mode(cap_entry);
+                    free(cap_entry->cap.kas_ecc_cap);
```
fliphil: Possible to move this free into acvp_cap_free_kas_ecc_mode?

src/acvp.c
```diff
@@ -4774,6 +4848,163 @@ static ACVP_RESULT acvp_build_dsa_register_cap (JSON_Object *cap_obj, ACVP_CAPS_
...
+                current_func = current_func->next;
+            }
+            json_object_set_value(cap_obj, "curve", json_value_init_array());
+            temp_arr = json_object_get_array(cap_obj, "curve");
+            current_curve = kas_ecc_mode->curve;
+            while (current_curve) {
+                switch (current_curve->param)
+                { 
+                case ACVP_ECDSA_CURVE_B233:
+                    json_array_append_string(temp_arr, "b-233");
```
fliphil: Replace strings with #defines?

Makefile.murl
bfussell: Yes, adding Makefile.win and acvp_kas_ecc.c(get fewer comments when omitted :)

app/app_main.c
bfussell: Spec change ? Never ! :)
Sure I'll move it inside.

src/acvp.h
bfussell: Unfortunately it does since the prereqs are in the mode(s) not the cap

app/app_main.c
bfussell: got it.

app/app_main.c
bfussell: I'm gonna need more info.

app/app_main.c
bfussell: got it.

app/app_main.c
bfussell: got it

src/acvp.c
bfussell: sure.

src/acvp.c
bfussell: pass for now - there so many other places. Maybe we could make a user story to move them all to #defines

PR #89, opened 2018-06-20
KAS-ECC Component mode and some info for non-Component mode

bfussell:
Plus fixed a bug in TOTP password generation when there was a NULL in the string which caused the key length to be wrong.

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -3559,10 +3559,11 @@ static ACVP_RESULT app_kas_ecc_handler(ACVP_TEST_CASE *test_case)
     int nid = 0, exout = 0;
     EC_KEY *ec = NULL;
     EC_POINT *peerkey = NULL;
-    unsigned char *Z;
-    int Zlen;
+    unsigned char *Z = NULL;
+    int Zlen = 0;
     BIGNUM *cx = NULL, *cy = NULL, *ix = NULL, *iy = NULL, *id = NULL;
     const EVP_MD *md = NULL;
+    ACVP_RESULT rv = ACVP_CRYPTO_MODULE_FAIL;
 
     tc = test_case->tc.kas_ecc;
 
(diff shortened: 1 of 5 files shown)


app/app_main.c
```diff
@@ -3645,23 +3670,60 @@ static ACVP_RESULT app_kas_ecc_handler(ACVP_TEST_CASE *test_case)
...
-        EC_GROUP_free(group);
-        BN_free(cx);
-        BN_free(cy);
-        printf("EC_KEY_generate_key failed\n");
-        return ACVP_CRYPTO_MODULE_FAIL;
+    if (tc->test_type == ACVP_KAS_ECC_TT_VAL) {
+        if (!BN_hex2bn(&ix, (char *)tc->pix)) {
+            EC_GROUP_free(group);
+            BN_free(cx);
+            BN_free(cy);
```
fliphil: Down to preference, but goto cleanup and then the memory can be freed there. Saves redundant lines for each for each of these error statements. Up to you :)

bfussell: I'll commit that change with the second half of the TOTP fix that I missed.
You'll have to re-approve I think.

app/app_main.c
```diff
@@ -3761,7 +3714,12 @@ static ACVP_RESULT app_kas_ecc_handler(ACVP_TEST_CASE *test_case)
...
     }
-    OPENSSL_cleanse(Z, Zlen);
+    rv = ACVP_SUCCESS;
+
+error:
+    if (Z) {
+        OPENSSL_cleanse(Z, Zlen);
+    }
     OPENSSL_free(Z);
     EC_KEY_free(ec);
```
fliphil: Do you need to check "if (ec) EC_KEY_free(ec);" ?

bfussell: no, only reason Z needed that check was since it was a cleanse and not a free

fliphil:
TOTP fix looks good 👍

PR #90, opened 2018-06-20
Add TOTP password refresh in response to HTTP 401

bfussell:
Also made results request endless since we now have a refresh on the password.

### DIFF
--- a/src/acvp_transport.c
+++ b/src/acvp_transport.c
@@ -38,6 +38,7 @@
 #include "acvp_lcl.h"
 
 #define HTTP_OK    200
+#define HTTP_UNAUTH    401
 
 #define MAX_TOKEN_LEN 600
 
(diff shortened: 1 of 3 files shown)


src/acvp_transport.c
```diff
@@ -595,9 +634,19 @@ ACVP_RESULT acvp_retrieve_vector_set_result (ACVP_CTX *ctx, int vs_id) {
     }
     rv = acvp_curl_http_get(ctx, url, &acvp_curl_write_kat_func);
     if (rv != HTTP_OK) {
-        ACVP_LOG_ERR("Unable to get vector result from server. curl rv=%d\n", rv);
-        ACVP_LOG_ERR("%s\n", ctx->kat_buf);
-        return ACVP_TRANSPORT_FAIL;
+        if (rv == HTTP_UNAUTH) {
```
fliphil: Since this code seems mostly common and used in 4 places, what do you think about making a new function "acvp_http_get" that encapsulates these return code checks and the refresh?

bfussell: I would except for some are GETs some are POSTs and there are 3 different write functions would be lots of parameters passed in.

PR #92, opened 2018-06-21
Kdf135 ssh library enhancements

fliphil:


src/acvp_kdf135_ssh.c
```diff
@@ -106,7 +114,7 @@ ACVP_RESULT acvp_kdf135_ssh_kat_handler (ACVP_CTX *ctx, JSON_Object *obj) {
     /*
      * Get the crypto module handler for this hash algorithm
      */
-    alg_id = acvp_lookup_cipher_index(alg_str);
+    alg_id = acvp_lookup_cipherwithmode_index(alg_str, mode_str);
```
bfussell: This new utility may be useful at some point, but since you're in the handler already  you already know that your alg_id is 
ACVP_KDF135_SSH correct ?  Some handler support multiple flavors of algorithms like AES-CBC, AES-CFB, etc. so they have to locate which one it is, I don't think the KDF135 handlers need to do that.

fliphil: Sure, I'll back out that utility and just set the alg_id

PR #95, opened 2018-06-26
Adds IKEv1 library support

briddavi:


fliphil:
Does it make sense to convert the hexstr to binary before passing to client? I don't see any changes around that.

PR #98, opened 2018-07-03
Cleanup 0629

elliemdaw:
- Splits acvp.c into multiple files
- cleans up some static analysis warnings
- fixes major cmac processing memory leak and a couple of other minor leaks


src/acvp_kdf108.c
```diff
@@ -94,13 +94,15 @@ static ACVP_RESULT acvp_kdf108_init_tc (ACVP_CTX *ctx,
                                         int deferred
 ) {
     memset(stc, 0x0, sizeof(ACVP_KDF108_TC));
+    ACVP_RESULT rv;
```
fliphil: Could you move this variable declaration above memset? Thanks!

fliphil:
Really like that you split up the functions into modules! Also nice cleanup job. My only request is that acvp_enable_cap.c be named to something like acvp_capabilities.c . If I were a new developer looking at the codebase, something like the second would be more clear as to what it contains IMO :)

PR #99, opened 2018-07-05
Support building app with openssl 1.1.0

fliphil:


app/app_main.c
```diff
@@ -2583,52 +2639,55 @@ static ACVP_RESULT app_aes_handler_aead(ACVP_TEST_CASE *test_case)
...
-            EVP_CIPHER_CTX_cleanup(&cipher_ctx);
-            return ACVP_NO_CAP;
+        rc = ACVP_NO_CAP;
+        goto end;
     }
 
-    EVP_CIPHER_CTX_cleanup(&cipher_ctx);
+end:
+    /* Cleanup */
+    if (cipher_ctx) EVP_CIPHER_CTX_free(cipher_ctx);
```
elliemdaw: Yes! Thank you... a similar thing was causing a big memory leak in the cmac code previously :)

app/app_main.c
```diff
@@ -130,12 +133,19 @@ char *ca_chain_file;
...
+static EVP_CIPHER_CTX *glb_cipher_ctx = NULL; /* need to maintain across calls for MCT */
 
 #define CHECK_ENABLE_CAP_RV(rv) \
     if (rv != ACVP_SUCCESS) { \
         printf("Failed to register capability with libacvp (rv=%d: %s)\n", rv, acvp_lookup_error_string(rv)); \
-        exit(1); \
+        return 1; \
+    }
+
+#define CHECK_ENABLE_CAP_RV_DEMO(rv) \
```
elliemdaw: Maybe I am missing it, where is this one used?

fliphil: Oops, this macro isn't needed here. It was intended for mirror. I'll remove

PR #106, opened 2018-07-30
Create version command line option to display library and protocol ve…

bfussell:
…rsion info

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -233,6 +233,7 @@ static void print_usage(int err)
     printf("      --status(default)\n");
     printf("      --info\n");
     printf("      --verbose\n");
+    printf("      --version\n");
     printf("\n");
     printf("Algorithm Test Suites:\n");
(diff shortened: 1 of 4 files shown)


app/app_main.c
```diff
@@ -354,6 +355,11 @@ static int ingest_cli(APP_CONFIG *cfg, int argc, char **argv) {
     argv++;
     argc--;
     while (argc >= 1) {
+        /* version option used by itself, ignore remaining command line */
+        if (strncmp(*argv, "--version", strlen("--version")) == 0) {
+            printf("\nACVP version %s(%s)\n", acvp_version(), acvp_protocol_version());
```
fliphil: From the cli ouput, a user might not know which number is the library vs protocol?

bfussell: Something like this better ?

ACVP library version(protocol version): 1.0.0(0.4)

PR #115, opened 2018-08-06
Prevent uninitialized return value in acvp.c

nescohen:
This pull request prevents an uninitialized return value in "acvp_check_test_results" in the case that the context contains no vector sets, hopefully preventing confusing error messages.
--
This issue was found by Muse, an automated code quality platform that uses advanced reasoning to find critical errors during development. Visit us at [musedev.io](http://musedev.io) to learn more and apply to join our private beta.

### DIFF
--- a/src/acvp.c
+++ b/src/acvp.c
@@ -1302,8 +1302,10 @@ ACVP_RESULT acvp_check_test_results (ACVP_CTX *ctx) {
      * in the regisration response.  Attempt to download the result
      * for each vector set.
      */
-	rv = ACVP_NO_CTX;
     vs_entry = ctx->vs_list;
+    if (!vs_entry) {
+        return ACVP_MISSING_ARG;
+    }
     while (vs_entry) {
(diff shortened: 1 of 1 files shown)


src/acvp.c
```diff
@@ -1302,6 +1302,7 @@ ACVP_RESULT acvp_check_test_results (ACVP_CTX *ctx) {
      * in the regisration response.  Attempt to download the result
      * for each vector set.
      */
+	rv = ACVP_NO_CTX;
```
elliemdaw: Maybe it would make more sense to add an additional error code for missing or invalid attributes? This error could be a bit misleading

nescohen: Yes, I was thinking along the lines of the comment saying "or no valid context", but I am new to this codebase. Adding a new error makes a lot of sense. I will make this change.

elliemdaw: The error codes are defined around line 1300 in acvp.h :) thank you!

PR #116, opened 2018-08-06
Prevent 2nd uninitialized return value in acvp.c

nescohen:
This pull request prevents an uninitialized return value in "acvp_process_tests" in the case that the context contains no vector sets, hopefully preventing confusing error messages.
--
This issue was found by Muse, an automated code quality platform that uses advanced reasoning to find critical errors during development. Visit us at [musedev.io](http://musedev.io) to learn more and apply to join our private beta.

### DIFF
--- a/src/acvp.c
+++ b/src/acvp.c
@@ -1255,8 +1255,10 @@ ACVP_RESULT acvp_process_tests (ACVP_CTX *ctx) {
      * in the regisration response.  Process each vector set and
      * return the results to the server.
      */
-	rv = ACVP_NO_CTX;
     vs_entry = ctx->vs_list;
+    if (!vs_entry) {
+        return ACVP_MISSING_ARG;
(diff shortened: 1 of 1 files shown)


src/acvp.c
```diff
@@ -1255,6 +1255,7 @@ ACVP_RESULT acvp_process_tests (ACVP_CTX *ctx) {
      * in the regisration response.  Process each vector set and
      * return the results to the server.
      */
+	rv = ACVP_NO_CTX;
```
elliemdaw: Similar to the other PR... maybe if there was an error code to handle invalid or missing args and then you can check for vs_entry == NULL instead? NO_CTX could be misleading

nescohen: That makes sense. I added the changes you suggested. Let me know if that looks good to you.

PR #131, opened 2018-08-23
More checks for kdf135_ikev2

fliphil:


letoams:
On Thu, 23 Aug 2018, Barry Fussell wrote:

> Subject: Re: [cisco/libacvp] More checks for kdf135_ikev2 (#131)

Related perhaps (I am not monitoring code changes too closely).

Up to now, NIST mistakenly adds SHA224 to its tests for IKEv1 and IKEv2.
But there is no IANA entry for SHA224. It is not a valid algorithm for
IKE. Please reject these in the code :)

Paul

fliphil:
Thanks Paul, good catch! If you open an issue at the following page, you can notify NIST directly so that they make the proper spec change: https://github.com/usnistgov/ACVP

Philip

PR #134, opened 2018-08-30
More checks for kdf135 ikev1

fliphil:


src/acvp_kdf135_ikev1.c
```diff
@@ -268,16 +290,87 @@ ACVP_RESULT acvp_kdf135_ikev1_kat_handler (ACVP_CTX *ctx, JSON_Object *obj) {
...
+        if (!auth_method_str) {
+            ACVP_LOG_ERR("Failed to include authenticationMethod");
+            return ACVP_MISSING_ARG;
+        }
+
+        /*
+         * Determine the authentication method.
+         */
+        if (strncmp(auth_method_str, ACVP_AUTH_METHOD_DSA_STR,
+                    strlen(ACVP_AUTH_METHOD_DSA_STR)) == 0) {
```
elliemdaw: strneln ? there are a few of these in this file :)

elliemdaw: Otherwise this all looks good to me :)

fliphil: Thanks! I'll double check the others in this file. Let me explain my reasoning behind using strlen here and see if you still want me to change it :)

Since the function is being used on a non-changing, hard-coded string (#define) we already know the length of the string and don't need to use strnlen. This is in contrast to user input where we don't have total control over the object and in that case need to verify the length with an upper bound in place to be safe. It's also good to also use strnlen for use with variables because the string value can also be expanded. However, since this string value is resolved by the pre-processor during compilation, it won't change during run-time.

@bfussell This is my current understanding. Am I making any incorrect assumptions here?

Thanks!
Philip

elliemdaw: I understand, and I thought about that too. I decided to comment because I think it would be good to just set precedence, you know? If we decide it isn't necessary then all good :) But I think it would be good to use those safe string functions everywhere if that is something we will care about in PRs in the future :)

bfussell: Apparently the other CSMs use strnlen even on hard coded strings, actually they use strnlen_s :)
In this case merge it as is for now. We'll need to go thru all the code to convert to the safeC APIs at some point, correct ?

PR #137, opened 2018-09-07
Adds more checks for kdf108

fliphil:
Also adds the previously missing "iv" field.


src/acvp_kdf108.c
```diff
@@ -104,16 +122,33 @@ static ACVP_RESULT acvp_kdf108_init_tc (ACVP_CTX *ctx,
     rv = acvp_hexstr_to_bin(key_in, stc->key_in, key_in_len, NULL);
     if (rv != ACVP_SUCCESS) return rv;
 
-    // Allocate space for the key_out
+    if (iv != NULL) {
+        /*
+         * Feedback mode.
```
elliemdaw: Should we double check mode here too?

src/acvp_kdf108.c
```diff
@@ -221,89 +275,126 @@ ACVP_RESULT acvp_kdf108_kat_handler (ACVP_CTX *ctx, JSON_Object *obj) {
...
+        if (strncmp(kdf_mode_str, ACVP_MODE_COUNTER,
+                    strlen(ACVP_MODE_COUNTER)) == 0) {
             kdf_mode = ACVP_KDF108_MODE_COUNTER;
-        } else if (strncmp(kdf_mode_str, ACVP_MODE_FEEDBACK, 64) == 0) {
+        } else if (strncmp(kdf_mode_str, ACVP_MODE_FEEDBACK,
+                           strlen(ACVP_MODE_FEEDBACK)) == 0) {
             kdf_mode = ACVP_KDF108_MODE_FEEDBACK;
-        } else if (strncmp(kdf_mode_str, ACVP_MODE_DPI, 64) == 0) {
+        } else if (strncmp(kdf_mode_str, ACVP_MODE_DPI,
+                           strlen(ACVP_MODE_DPI)) == 0) {
```
elliemdaw: (strlen --> strnlen) I assume these will be part of this later user story to sweep through all these? I think most of them were converted when safec was integrated, but I guess we will need another pass at some point

src/acvp_kdf108.c
fliphil: Don't think we need to since the other function does the check (which you already know) :) If the design changes in the future we can adapt as needed

src/acvp_kdf108.c
fliphil: Yeah I think doing all of them in a dedicated story is best

PR #173, opened 2018-11-12
API changes incoming...

fliphil:
* Normalized all APIS to use only enumeration values to represent "types" (still some strings when appropriate).
* Only use enum value for hash alg in public structs
* Changed API function names for clarity.
* Reworked some private functions dealing with appending objects to acvp capability list
* New stuff in the utilities.


src/acvp_build_register.c
```diff
@@ -2003,40 +1972,40 @@ static ACVP_RESULT acvp_build_kas_ecc_register_cap (ACVP_CTX *ctx, JSON_Object *
                     curve = current_pset->curve;
                     switch (curve)
                     {
-                    case ACVP_ECDSA_CURVE_B233:
+                    case ACVP_EC_CURVE_B233:
```
elliemdaw: Could this switch be replaced with acvp_lookup_ec_curve_name() too?

fliphil: yes it sure can be! I'll make the change

PR #187, opened 2018-11-28
Interop for AES/DES CTR mode

bfussell:
Will add UT after merging to internal


app/app_main.c
```diff
@@ -2786,6 +2791,9 @@ static ACVP_RESULT app_des_handler(ACVP_TEST_CASE *test_case) {
         cipher = EVP_des_ede3_cfb1();
         break;
     case ACVP_TDES_CTR:
+        iv = tc->iv;
+        cipher = EVP_des_ede3_ctr();
+        break;
```
elliemdaw: I think we left this out because it won't build out-of-the-box with openssl 1.0.2. Do we want it included?

bfussell: Yep, good catch. I'll add it to the internal repo after the next merge.

PR #189, opened 2018-12-02
Changes return values for application level crypto handlers

elliemdaw:


bfussell:
I don't think we ever documented the return code options for the crypto_handler and since we're using an int now we should probably document that somewhere for those that are adding new algorithms or all new crypto modules. Someone may ASSUME return 1 indicated success :)

PR #230, opened 2019-02-12
Adds unit tests to external repo

elliemdaw:
Preparing for IETF, it will help build an open source community around this library if contributors were able to test their contributions.

This PR adds the existing UTs to the external library.


fliphil:
There are some .pyc files that should be removed.

elliemdaw:
Ahh good catch

PR #232, opened 2019-02-15
Refactor transport module

fliphil:


TomMD:
Oops, some spurious messages here obviously (invalid dead store warnings).  These shouldn't appear again and the issue preventing all comments has been addressed.

PR #239, opened 2019-03-14
Add domain for symmetric cipher registration

fliphil:


fliphil:
@bfussell Hmm looks like maybe the server doesn't support this right now. I'll keep around for now but hold off on merge I guess

fadaile:
dear all
donat

PR #240, opened 2019-03-15
add command line option to read JSON vector file and process as if it…

bfussell:
… came from server

Should have thought of this sooner :(
Should help with the GCM-SIV stuff as well as testing on ASA
Best used with --verbose mode on


fliphil:
So is this intended to be run in something like gdb, and then inspect the fields after they have run through the "process" step by the crypto module?

PR #248, opened 2019-04-09
Acvp 1.0 support all algorithms

fliphil:


src/acvp_build_register.c
```diff
@@ -105,13 +105,16 @@ static ACVP_RESULT acvp_build_hash_register_cap(JSON_Object *cap_obj, ACVP_CAPS_
...
+    const char *revision = NULL;
 
     if (!hash_cap) {
         return ACVP_MISSING_ARG;
     }
 
     json_object_set_string(cap_obj, "algorithm", acvp_lookup_cipher_name(cap_entry->cipher));
-    json_object_set_string(cap_obj, "revision", ACVP_SHA_REVISION);
+
+    revision = acvp_lookup_cipher_revision(cap_entry->cipher);
```
bfussell: Do we want to generate some type of error generated when we get a NULL ?
Maybe just do it in the lookup function ?

fliphil: Ah good catch

PR #254, opened 2019-04-12
address a couple of the muse errors

bfussell:


fliphil:
You might want to merge in master because I changed that token stuff with a recent commit

PR #257, opened 2019-04-18
Sha3 and Shake

fliphil:
* Also adds support for large/ endpoint


bfussell:
I guess we'll need some UT, but we can make a separate user story for that.  Does the old hash UT still work with no leaks ?

src/acvp.c
```diff
@@ -1476,6 +1491,102 @@ static ACVP_RESULT acvp_parse_login(ACVP_CTX *ctx) {
...
+    val = json_value_init_object();
+    obj = json_value_get_object(val);
+
+    /* 
+     * Cut off the https://name:port/ prefix and /results suffix
+     */
+    strstr_s((char *)url, ACVP_ATTR_URL_MAX, "/acvp/v1", 8, &substr);
+    strcpy_s(snipped_url, ACVP_ATTR_URL_MAX, substr);
+    strstr_s(snipped_url, ACVP_ATTR_URL_MAX, "/results", 8, &substr);
+    if (!substr) return ACVP_INVALID_ARG;
```
bfussell: goto err ? wouldn't arr_val be leaked otherwise ?

src/acvp_hash.c
```diff
@@ -183,6 +199,226 @@ static ACVP_RESULT acvp_hash_mct_tc(ACVP_CTX *ctx,
...
+    unsigned int range = max_xof_bytes - min_xof_bytes + 1;
+
+    /*
+     * Initial Outputlen = (floor(maxoutlen/8) )*8
+     */
+    xof_len = (max_xof_bits / 8) * 8;
+    /* Convert from bits to bytes */
+    stc->xof_len = (xof_len + 7) / 8;
+
+#if 0
```
bfussell: Still needed or something we need to look at later ?

src/acvp_hash.c
```diff
@@ -183,6 +199,226 @@ static ACVP_RESULT acvp_hash_mct_tc(ACVP_CTX *ctx,
...
+                ACVP_LOG_ERR("crypto module failed the operation");
+                rv = ACVP_CRYPTO_MODULE_FAIL;
+                goto end;
+            }
+
+            /* Get the right-most 16bits and convert to an integer */
+#ifdef WIN32
+            rightmost_out_bits = _byteswap_ushort(*(uint16_t *)(stc->md + stc->md_len - 2));
+#else
+# if ACVP_HOST_LITTLE_ENDIAN
```
bfussell: Got to be a 16 bit byte swap libc API that is endian agnostic I would think. Maybe not that supports windows and Mac ?

fliphil: Do you have any specific in mind? I think ntohs is also endian dependent

PR #262, opened 2019-04-23
remove extra / on some URI strings

bfussell:
Those that use path_segment need the / when building the URI, these that were changed do not.

### DIFF
--- a/src/acvp_transport.c
+++ b/src/acvp_transport.c
@@ -471,7 +471,7 @@ ACVP_RESULT acvp_submit_vector_responses(ACVP_CTX *ctx, char *vsid_url) {
     }
 
     snprintf(url, ACVP_ATTR_URL_MAX - 1,
-            "https://%s:%d/%s/results",
+            "https://%s:%d%s/results",
             ctx->server_name, ctx->server_port, vsid_url);
 
     return acvp_network_action(ctx, ACVP_NET_POST_VS_RESP, url, NULL, 0);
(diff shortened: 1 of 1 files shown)


fliphil:
Thanks for fixing this

src/acvp_transport.c
```diff
@@ -577,7 +577,7 @@ ACVP_RESULT acvp_transport_get(ACVP_CTX *ctx, const char *url) {
     }
 
     snprintf(constructed_url, ACVP_ATTR_URL_MAX - 1,
-             "https://%s:%d/%s",
+             "https://%s:%d%s",
```
fliphil: I think sometimes the string here might not have the double slash. Wondering if we should do a strstr looking for "//" instead. If that's found then overwrite one of the char

bfussell: Are you referring to all instances or just this one ?
This one isn't even used, in fact I started to rip that function out but figured it may be something we could use in the future

fliphil: Thinking inside of acvp_network_action

bfussell: Although acvp_network_action is is common to all the operations it seems a bit problematic adding sanitation there since that means you have to scan the URL and possibly shrink the string. There are certain times we must expect the / to be the start of the string - everything the server gives us will be that way.  The others are all built by our code so I think we can simply make sure no sanitation is needed for those.

fliphil: Yeah, there would need to be a new url[] buffer in acvp_network_action. There's 2 things that make me think we might want to scan/sanitize.
1. The spec doesn't specify whether the url starts with / or not. So some other servers may choose to omit it? We want the client to be robust against that potential variability.
2. I wonder if doing a sanity check will be helpful with more complicated URL that contains GET query parameters (or other features of a URL that we want to check).

For URL that we build entirely from scratch this scan/sanitize approach is useless. It's useful for URLs that have portions not made by us though (i.e. given from server).
ATM your approach will get the job done.
Let's think about whether we want to add a feature like this though before first release :)

PR #263, opened 2019-04-25
Operating environment metadata parsing

fliphil:


src/acvp_operating_env.c
```diff
@@ -0,0 +1,1684 @@
+/** @file */
+/*****************************************************************************
+* Copyright (c) 2019, Cisco Systems, Inc.
```
bfussell: We're using the Apache 2.0 license now ;)

src/acvp_build_register.c
```diff
@@ -2810,113 +2809,204 @@ ACVP_RESULT acvp_build_oes(ACVP_CTX *ctx, char **reg) {
...
+    }
+
+    if (vendor->email) {
+        json_object_set_value(obj, "emails", json_value_init_array());
+        email_array = json_object_get_array(obj, "emails");
+
+        json_array_append_string(email_array, vendor->email);
     }
 
+#if 0
```
bfussell: waiting on support for multiple phone numbers ?

fliphil: This stuff that deffed out was originally for doing the creation of the resources. Kept it around for when we end up doing the query, I'll tweak the code here

PR #273, opened 2019-05-30
Metadata.py add query syntax

fliphil:


bfussell:
At some point we need real Cisco infor in the internal repo to help jump start internal customers

fliphil:
@bfussell Roger that, I'll make it a priority once it's in the repo

PR #274, opened 2019-05-31
offline mode for processing vectors

bfussell:
and a few cleanup items

Have not removed cURL from build, mURL can be used.
will create a spike to eval removing cURL.

steps need to be added to README  with decriptions -will create user story for that.
they are:

app/.libs/acvp_app --kas_ffc --vector_req req.json --none
app/.libs/acvp_app --all_algs --vector_req req.json  --vector_rsp rsp.json
app/.libs/acvp_app --all_algs --vector_upload rsp.json

### DIFF
--- a/include/acvp/acvp.h
+++ b/include/acvp/acvp.h
@@ -2513,6 +2513,17 @@ ACVP_RESULT acvp_set_certkey(ACVP_CTX *ctx, char *cert_file, char *key_file);
  */
 ACVP_RESULT acvp_mark_as_sample(ACVP_CTX *ctx);
 
+/*! @brief acvp_mark_as_request_only() marks the registration as a request only.
+
+    This function sets a flag that will allow the client to retrieve
+    the vectors from the server and store them in a file for later use.
+
+    @param ctx Pointer to ACVP_CTX that was previously created by
+        calling acvp_create_test_session.
+ */
+ACVP_RESULT acvp_mark_as_request_only(ACVP_CTX *ctx, char *filename);
+
+
 /*! @brief Performs the ACVP testing procedures.
  *
  * This function will do the following actions:
@@ -2585,6 +2596,8 @@ ACVP_RESULT acvp_set_json_filename(ACVP_CTX *ctx, const char *json_filename);
  * @return ACVP_RESULT
  */
 ACVP_RESULT acvp_load_kat_filename(ACVP_CTX *ctx, const char *kat_filename);
+ACVP_RESULT acvp_upload_vectors_from_file(ACVP_CTX *ctx, const char *rsp_filename);
+ACVP_RESULT acvp_run_vectors_from_file(ACVP_CTX *ctx, const char *req_filename, const char *rsp_filename);
 
 /*! @brief acvp_set_2fa_callback() sets a callback function which
     will create or obtain a TOTP password for the second part of
--- a/include/acvp/acvp.h
+++ b/include/acvp/acvp.h
@@ -2520,6 +2520,8 @@ ACVP_RESULT acvp_mark_as_sample(ACVP_CTX *ctx);
 
     @param ctx Pointer to ACVP_CTX that was previously created by
         calling acvp_create_test_session.
+    @param filename Name of the file to be used for the request vectors
+
  */
 ACVP_RESULT acvp_mark_as_request_only(ACVP_CTX *ctx, char *filename);
 
--- a/include/acvp/acvp_lcl.h
+++ b/include/acvp/acvp_lcl.h
@@ -1253,6 +1253,9 @@ struct acvp_ctx_t {
     char *json_filename;
     int use_json;
     int is_sample;
+    char *vector_req_file;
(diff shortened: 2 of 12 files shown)


include/acvp/acvp.h
```diff
@@ -2513,6 +2513,17 @@ ACVP_RESULT acvp_set_certkey(ACVP_CTX *ctx, char *cert_file, char *key_file);
  */
 ACVP_RESULT acvp_mark_as_sample(ACVP_CTX *ctx);
 
+/*! @brief acvp_mark_as_request_only() marks the registration as a request only.
+
+    This function sets a flag that will allow the client to retrieve
+    the vectors from the server and store them in a file for later use.
+
+    @param ctx Pointer to ACVP_CTX that was previously created by
+        calling acvp_create_test_session.
```
fliphil: Missing param for filename here

include/acvp/acvp_lcl.h
```diff
@@ -1253,6 +1253,9 @@ struct acvp_ctx_t {
     char *json_filename;
     int use_json;
     int is_sample;
+    char *vector_req_file;
```
fliphil: Comments for these fields may help :)

src/parson.c
```diff
@@ -1650,6 +1650,65 @@ JSON_Status json_serialize_to_file_pretty(const JSON_Value *value, const char *f
...
 }
 
+JSON_Status json_serialize_to_file_pretty_a(const JSON_Value *value, const char *filename) {
+    JSON_Status return_code = JSONSuccess;
+    FILE *fp = NULL;
+    char *serialized_string = NULL; 
+
+    fp = fopen(filename, "a");
+    if (fp == NULL) {
+        json_free_serialized_string(serialized_string);
```
fliphil: Dont need to free the string here right?

src/parson.c
```diff
@@ -1650,6 +1650,65 @@ JSON_Status json_serialize_to_file_pretty(const JSON_Value *value, const char *f
...
+    json_free_serialized_string(serialized_string);
+    return return_code;
+}
+
+JSON_Status json_serialize_to_file_pretty_w(const JSON_Value *value, const char *filename) {
+    JSON_Status return_code = JSONSuccess;
+    FILE *fp = NULL;
+    char *serialized_string = json_serialize_to_string_pretty(value, NULL);
+    if (serialized_string == NULL) {
+        return JSONFailure;
```
fliphil: Do need to free the string here before return

src/acvp.c
```diff
@@ -803,6 +807,334 @@ ACVP_RESULT acvp_load_kat_filename(ACVP_CTX *ctx, const char *kat_filename) {
...
+        json_result = json_serialize_to_string_pretty(kat_val, NULL);
+        file_val = json_parse_string(json_result);
+        json_free_serialized_string(json_result);
+
+        /* track first vector set with file count */
+        if (n == 1) {
+            ACVP_STRING_LIST *vs_entry = NULL;
+
+            rsp_val = json_value_init_object();
+            rsp_obj = json_value_get_object(rsp_val);
```
fliphil: Can you not use the val/obj that you parsed earlier from the req file? Instead of making new ones here

src/acvp.c
```diff
@@ -1528,15 +1891,52 @@ static ACVP_RESULT acvp_process_vsid(ACVP_CTX *ctx, char *vsid_url) {
...
+
+            /*
+             * Save the KAT VectorSet to file
+             */
+            if (ctx->vector_req) {
+                alg_array = json_value_get_array(val);
+                alg_val = json_array_get_value(alg_array, 1);
+
+                /* track first vector set with file count */
+                if (count == 0) {
```
fliphil: What is count being used for?

include/acvp/acvp_lcl.h
bfussell: sure

src/parson.c
bfussell: correct, will remove

include/acvp/acvp.h
bfussell: correct will add

src/acvp.c
bfussell: I am trying to be very explicit there since there are several values used in this function

src/parson.c
```diff
@@ -1650,6 +1650,65 @@ JSON_Status json_serialize_to_file_pretty(const JSON_Value *value, const char *f
...
+        }
+    }
+    if (fclose(fp) == EOF) {
+        return_code = JSONFailure;
+    }
+    json_free_serialized_string(serialized_string);
+    return return_code;
+}
+
+JSON_Status json_serialize_to_file_pretty_w(const JSON_Value *value, const char *filename) {
```
fliphil: Unless you are using private functions here, might want to move these functions to utils.c instead. To try and keep this file as sane as possible for future syncs from the official

bfussell: agree will do

PR #282, opened 2019-06-13
Re-enable SigGen and address missing 00 at start of signature

bfussell:
This is merely a single workaround for a bigger problem documented as an issue on the ACVP github.
If they don't fix it and there are other instances, we will need to create a common utility to pad strings when they are not fully qualified.

### DIFF
--- a/src/acvp_rsa_sig.c
+++ b/src/acvp_rsa_sig.c
@@ -200,11 +200,11 @@ static ACVP_RESULT acvp_rsa_sig_kat_handler_internal(ACVP_CTX *ctx, JSON_Object 
 
     ACVP_CIPHER alg_id;
     char *json_result = NULL, *mode_str;
-    unsigned int mod = 0;
-    char *msg, *signature = NULL;
+    unsigned int mod = 0, padding = 0;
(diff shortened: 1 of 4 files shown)


src/acvp_rsa_sig.c
```diff
@@ -399,26 +399,42 @@ static ACVP_RESULT acvp_rsa_sig_kat_handler_internal(ACVP_CTX *ctx, JSON_Object
...
+                json_siglen = strnlen_s(tmp_signature, ACVP_RSA_SIGNATURE_MAX + 1);
                 if (json_siglen > ACVP_RSA_SIGNATURE_MAX) {
                     ACVP_LOG_ERR("'signature' too long in server json");
                     rv = ACVP_INVALID_ARG;
                     json_value_free(r_tval);
                     goto err;
                 }
+
+                if (json_siglen != mod/4) {
+                    signature = calloc(mod/4 +1, sizeof(char));
```
fliphil: Need to check is calloc was successful here?

bfussell: yep.

PR #284, opened 2019-06-17
fips validation PUT, and metadata querying

fliphil:


app/app_lcl.h
```diff
@@ -22,7 +22,7 @@ extern "C"
  */
 #define DEFAULT_SERVER "127.0.0.1"
 #define DEFAULT_PORT 443
-#define DEFAULT_URI_PREFIX "acvp/v1/"
+#define DEFAULT_URI_PREFIX "/acvp/v1/"
```
bfussell: Wasn't this changed because of the double slashes ?

metadata/validation.json
```diff
@@ -1,32 +1,33 @@
...
+                "country": "USA",
+                "postalCode": "00000"
+            },
             "contacts": [
                 {
-                    "full_name": "Wyle E. Coyote",
+                    "fullName": "Jane Smith",
                     "emails": [
-                        "wcoyote@acme-fictional.com"
+                        "jane.smith@acme.acme"
```
bfussell: acme.acme ? is that supposed to be acme.com ?

app/app_lcl.h
fliphil: Hey I changed this because every URL the serve hands us starts with /acvp/v1/ and the new acvp_transport_get function I made to handle constructed strings along with the ones the server hands us. It was easier to "fix" this default prefix to have / leading

metadata/validation.json
fliphil: yeah that was a typo, but it worked for the creation and query

PR #290, opened 2019-06-26
Utilize the new PUT flow for alg cert request in offline mode

bfussell:
and fix a leak triggered by retries

### DIFF
--- a/src/acvp.c
+++ b/src/acvp.c
@@ -25,6 +25,10 @@
 /*
  * Forward prototypes for local functions
  */
+static ACVP_RESULT acvp_validate_test_session(ACVP_CTX *ctx);
+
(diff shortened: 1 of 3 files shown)


src/acvp.c
```diff
@@ -1055,6 +1059,18 @@ ACVP_RESULT acvp_upload_vectors_from_file(ACVP_CTX *ctx, const char *rsp_filenam
         goto end;
     }
 
+    if (fips_validation) {
+        rv = fips_metadata_ready(ctx);
+        if (ACVP_SUCCESS != rv) {
+            ACVP_LOG_ERR("Validation metadata not ready");
+            return ACVP_UNSUPPORTED_OP;
```
fliphil: This needs to be goto end I think

bfussell: yep

PR #294, opened 2019-07-23
Fix linker option for noexecstack

bfussell:


bfussell:
Zero length PT will now be supported by new algorithm called AES-GMAC.

PR #299, opened 2019-08-09
include libcrypto in static build of roffline mode

bfussell:


configure.ac
```diff
@@ -160,6 +160,7 @@ AS_IF(
  [test "x$enable_offline" != "xfalse"],
  [[CFLAGS="$CFLAGS -DACVP_OFFLINE"]
   [LDFLAGS="-static $LDFLAGS"]
+  [CC="$fomdir/bin/fipsld"]
```
fliphil: Might want to add $CC in there in case the user passed in more externally

bfussell: Normally yes, but in the case of using fipsld for compilation there can be no other options included.

PR #306, opened 2019-08-21
Integrated AES-GMAC support into the library and sample app

abkarcher:
- Integrated AES-GMAC support into library, basically AES-GCM with pt/ct length of 0
- Added AES-GMAC registration and handling into sample application


src/acvp_aes.c
```diff
@@ -810,6 +810,11 @@ ACVP_RESULT acvp_aes_kat_handler(ACVP_CTX *ctx, JSON_Object *obj) {
         }
 
         ptlen = (unsigned int)json_object_get_number(groupobj, "payloadLen");
+        if(alg_id == ACVP_AES_GMAC && ptlen != 0) {
```
bfussell: nit. Your if statements should always have a space between if and (
There are few more cases in the diff.

PR #311, opened 2019-08-30
Feature/aes gcm siv

abkarcher:
-Completed library implementation of aes-gcm-siv.
-Currently not supported by OpenSSL, unable to implement in sample app
-Not currently tested with ACVP server due to lack of openSSL support
-Tested to build with 1.0.2 and 1.1.1, and to run other AES tests without issue

### DIFF
--- a/src/acvp_aes.c
+++ b/src/acvp_aes.c
@@ -733,11 +733,17 @@ ACVP_RESULT acvp_aes_kat_handler(ACVP_CTX *ctx, JSON_Object *obj) {
             (alg_id != ACVP_AES_KWP)) {
             ivlen = 128;
         }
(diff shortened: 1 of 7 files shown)


src/acvp_aes.c
```diff
@@ -790,6 +791,12 @@ ACVP_RESULT acvp_aes_kat_handler(ACVP_CTX *ctx, JSON_Object *obj) {
                     rv = ACVP_INVALID_ARG;
                     goto err;
                 }
+            } else if (alg_id == ACVP_AES_GCM_SIV) {
+                if (ivlen != 96) {
```
bfussell: I guess we should make a #define for the GCM-SIV IV length

abkarcher: And tag length - I will get on that

PR #314, opened 2019-09-03
Server Request Retry Backoff

abkarcher:
- All instances of waiting for server to create or process vector sets now have a maximum wait time (1 hour for now)
- There is now a retry backoff for waiting for the server to complete processing at vector sets (starts at 30 seconds, doubles until it reaches max of 5 minutes, easy to change as needed)
- Tested w/ NIST server by using arbitrarily low retry and max wait times
- Note: There is no retry backoff for waiting on the server to CREATE vector sets currently, as that retry time is provided by the server as per the spec
- Note: Fixed bug where error return values from acvp_process_vsid would not be handled properly if acvp_json_serialize_to_file_pretty_a (called afterwards) had no errors
- Note: todo: create sample app solution for reconnecting and retrieving vector set results at a later time

### DIFF
--- a/src/acvp.c
+++ b/src/acvp.c
@@ -2059,8 +2059,7 @@ static ACVP_RESULT acvp_process_vsid(ACVP_CTX *ctx, char *vsid_url, int count) {
             /*
              * Wait and try again to retrieve the VectorSet
              */
-            if (acvp_retry_handler(ctx, (unsigned int *)&retry_period, (unsigned int *)&time_waited_so_far, 1)
-                    != ACVP_KAT_DOWNLOAD_RETRY) {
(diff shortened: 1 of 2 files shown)


src/acvp.c
```diff
@@ -2032,7 +2059,12 @@ static ACVP_RESULT acvp_process_vsid(ACVP_CTX *ctx, char *vsid_url, int count) {
             /*
              * Wait and try again to retrieve the VectorSet
              */
-            acvp_retry_handler(ctx, retry_period);
+            if (acvp_retry_handler(ctx, (unsigned int *)&retry_period, (unsigned int *)&time_waited_so_far, 1)
```
bfussell: Do these two need to be type cast ? Seems like just using & should work since they are defined as unsigned int.
Same below where acvp_retry_handler is called as well.

PR #317, opened 2019-09-24
HTTP User-Agent Upgrade

abkarcher:
The HTTP User-agent now consists of: 

Library name/version; OS name; os/kernel version; processor architecture; processor name; compiler/version.

On Linux, Windows, and Mac OS on x86 or x64 platforms, this is all automatically populated. In other situations, depending on the environment, some of this information may be populated. For any information that cannot be found, an environment variable can be defined (undefined variables for info not automatically detected by the library throw a warning but do not prevent user from continuing).

Tested on Ubuntu, Windows 10, and Mac OS.

The Safe C Stub now contains function strremovews_s, which allows for leading and trailing white space on strings to be safely removed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,8 +66,9 @@ model {
 				linker.args "-lcurl"
 			} else {
 				cCompiler.define "WIN32"
-                cCompiler.define "__STDC_WANT_SECURE_LIB__=0"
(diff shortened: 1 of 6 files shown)


TomMD:
Sorry for the delay on that analysis.  They might be false positives to boot. I'm investigating both issues.

PR #318, opened 2019-10-08
Additional compiler option first commit for src and app directories

bfussell:
Will do test/ and app/ next and then turn on add compiler options for default build.

### DIFF
--- a/app/app_utils.c
+++ b/app/app_utils.c
@@ -92,6 +92,9 @@ base64_decode(const char *in, unsigned int inlen, unsigned char *out) {
             continue;
         case 3:
             out[j++] += (unsigned char)c;
+            continue;;
+        default:
+            return 0;
         }
     }
 
@@ -144,7 +147,6 @@ static ACVP_RESULT totp(char **token, int token_max) {
     char hash[MAX_LEN] = {0};
     int os, bin, otp;
     int md_len;
-    char format[5];
     time_t t;
     unsigned char token_buff[T_LEN + 1] = {0};
     char *new_seed = NULL;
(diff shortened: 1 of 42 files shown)


src/parson.c
```diff
@@ -488,10 +490,16 @@ static JSON_Status json_object_dotremove_internal(JSON_Object *object, const cha
     JSON_Object *temp_object = NULL;
     char *dot_pos = NULL;
     int name_len = 0;
+    char *name2 = NULL;
 
     /* SAFEC modified to use _s */
     name_len = strnlen_s(name, STRING_NAME_MAX);
-    strstr_s((char *)name, name_len, ".", 1, &dot_pos);
+    name2 = calloc(STRING_NAME_MAX, sizeof(char));
```
abkarcher: Unsure I follow, so just double checking:
- is name2 ever filled with data? seems to be an empty buffer for its lifespan
-Is it possible to lose a char of data if name_len = STRING_NAME_MAX?  (aka should calloc size be STRING_NAME_MAX + 1)

bfussell: Good catch, didn't cut/paste the strncpy_s to each instance, will add that and re-test.

app/app_utils.c
```diff
@@ -201,10 +203,12 @@ static ACVP_RESULT totp(char **token, int token_max) {
 
     otp = bin % DIGITS_POWER[ACVP_TOTP_LENGTH];
 
-    // generate format string like "%06d" to fix digits using 0
-    sprintf(format, "%c0%ldd", '%', (long int)ACVP_TOTP_LENGTH);
-
-    sprintf((char *)token_buff, format, otp);
+    /* If we ever change the size then it will fail hard here */
+    if (ACVP_TOTP_LENGTH != 8) {
```
abkarcher: Is this something we may want to make a #define for?

bfussell: Will address in next PR, thanks

PR #365, opened 2020-03-12
Get murl working better

bfussell:


src/acvp_transport.c
```diff
@@ -1162,6 +1163,7 @@ ACVP_RESULT acvp_transport_get(ACVP_CTX *ctx,
         }
 #ifndef USE_MURL
         while (1) {
+            const ACVP_KV_LIST *param = parameters;
```
abkarcher: Would this have undefined behavior if parameters = NULL?

bfussell: Will add a input arg check

bfussell: Actually there is already a check for parameters == NULL above this so we don't get here if it is.

PR #375, opened 2020-03-23
Fixed buffer overflow issue in old code 

abkarcher:
Some faulty logic allowed writes past edge of buffer indefinitely.
(I had not altered or looked in-depth at this code previously)

Code will now just cut off any characters longer than the buffer instead of overflowing.


abkarcher:
It was a "dead store" in the sense that it stored a value it never used before being updated, just replaced it with 0

PR #391, opened 2020-04-28
Process additional dependency fields

bfussell:


include/acvp/acvp_lcl.h
```diff
@@ -13,7 +13,7 @@
 #include "parson.h"
 
 #define ACVP_VERSION    "1.0"
-#define ACVP_LIBRARY_VERSION    "libacvp_oss-1.0.0"
+#define ACVP_LIBRARY_VERSION    "libacvp_oss-1.0.1"
```
abkarcher, line 16: May want to update micro version in configure.ac as well, though that can be done in a later PR if we are in a hurry

bfussell: I'll update it now

PR #420, opened 2020-06-08
New Windows Build System using Visual Studio instead of Gradle

abkarcher:
**LARGE PR**

See README.md for more information.

Tested in all configs for both app and library in x64.

Any suggestions/feedback appreciated.


abkarcher:
> 
> 
> Anything special need to done to the jenkins jobs to build and run UT ?

There currently isnt a system in place to build UT (There was not before so was unsure if there was need) - I'll happily work on this if need be

bfussell:
Build should suffice, UT is only run on linux and that should be good enough.

abkarcher:
> 
> 
> Build should suffice, UT is only run on linux and that should be good enough.

Ill look into it, thanks!

PR #431, opened 2020-06-19
Updated refresh check

abkarcher:
Should investigate if there is a better way to handle this scenario in the future, or maybe if a expired JWT message should be added to the spec.

Tested by setting retry values to 30 minutes, hence the delay in getting this in.


bfussell:
The format of the expired JWT message should probably be in the spec so it is defined and therefore should be enough to prevent them from changing it.

PR #455, opened 2020-07-24
Enhance drbg testing to include all combinations or PR and Reseed opt…

bfussell:
…ions

### DIFF
--- a/src/acvp_drbg.c
+++ b/src/acvp_drbg.c
@@ -353,8 +353,12 @@ ACVP_RESULT acvp_drbg_kat_handler(ACVP_CTX *ctx, JSON_Object *obj) {
 
                 /* Get 1st element from the array */
                 pr_input_val = json_array_get_value(pred_resist_input, index);
+                if (pr_input_val == NULL) {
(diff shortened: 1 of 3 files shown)


src/acvp_drbg.c
```diff
@@ -335,76 +347,148 @@ ACVP_RESULT acvp_drbg_kat_handler(ACVP_CTX *ctx, JSON_Object *obj) {
                 goto err;
             }
 
-            ACVP_LOG_VERBOSE("Found new DRBG Prediction Input...");
+            index = 0;
+            if (!pred_resist_enabled && reseed) {
+                ACVP_LOG_VERBOSE("Found new DRBG Prediction Input...");
+
+                /* Get 1st element from the array */
+                pr_input_val = json_array_get_value(pred_resist_input, index);
```
abkarcher, line 355: Maybe redundant because of the pr_input_count check, but do we want to check to make sure pr_input_val or pr_input_obj != null? Otherwise we get down into a strncmp_s without being certain if the string exists, unsure if that is a problem. Two other occurances of this below. if thats not an issue everything looks good to me.

bfussell: Added

PR #492, opened 2020-09-14
 Update revision for RSA and version number (1.0.6)

abkarcher:


bfussell (approved):

mtdownz:
This update may negatively impact ECDSA where you have a potential bug in src/acvp.c lines 135 to 137
image

abkarcher:
> 
> 
> This update may negatively impact ECDSA where you have a potential bug in src/acvp.c lines 135 to 137
> image

Thanks for the catch! We had fixed this in 1.1.X but that fix never made it into 1.0.X. Will fix before release. Thanks!

PR #502, opened 2020-09-24
Added library support for aes-xpn and disabled app registration

abkarcher:
Tested with a sample session and zereod out results (got "fail" disposition but all fields and data were present)

Fixed small issue where negative value could be fed to some sym cipher parm API. Fortunately only caused an error from the server.


bfussell (approved):
Great stuff !

PR #528, opened 2020-11-13
First commit of KAS-IFC-SSC

bfussell:
UT still in progress

### DIFF
--- a/include/acvp/acvp_lcl.h
+++ b/include/acvp/acvp_lcl.h
@@ -93,7 +93,7 @@
 #define ACVP_LOG_TRUNCATED_STR "...[truncated]\n"
 //This MUST be the length of the above screen (want to avoid calculating at runtime frequently)
 #define ACVP_LOG_TRUNCATED_STR_LEN 15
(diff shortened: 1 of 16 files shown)


include/acvp/acvp_lcl.h
```diff
@@ -93,7 +93,7 @@
 #define ACVP_LOG_TRUNCATED_STR "...[truncated]\n"
 //This MUST be the length of the above screen (want to avoid calculating at runtime frequently)
 #define ACVP_LOG_TRUNCATED_STR_LEN 15
-#define ACVP_LOG_MAX_MSG_LEN 2048
+#define ACVP_LOG_MAX_MSG_LEN 1024000
```
abkarcher, line 96: Was this something we wanted to keep? Will make info and verbose logging extremely long with showing 1MB of every vector set that comes through

abkarcher (approved):

include/acvp/acvp_lcl.h
bfussell: Nope, just fixed it.

PR #532, opened 2020-11-17
Fixing len for snmp

abkarcher:
p_len changed from bytes to bits as per https://github.com/usnistgov/ACVP/issues/1051


bfussell (approved):
On production they can use the previous version until it is updated ?

abkarcher:
> 
> 
> On production they can use the previous version until it is updated ?

yep, I had mentioned this as a solution to someone. Thanks!

PR #559, opened 2021-01-21
Fix KAS-IFC leaks and remove unused function setup_fail

bfussell:
This may conflict since I already added it to internal, so I'll look out for merge conflict on the sync.


abkarcher (approved):

abkarcher:
I have a few more things to push external that are already internal once the pipeline is complete too

PR #561, opened 2021-01-27
Fixing criterion testing issues

abkarcher:
The sandbox-gmon files were actually profile data files; when I revamped the build system a bit -p found its way into the test make data.

This was also causing valgrind to throw some harmless warnings during runs.

We still use debugging symbols with -g though.


bfussell (approved):
Great find !

PR #569, opened 2021-02-08
Add UT for new FFC cases and add remaining MODP support

bfussell:


abkarcher:
Can change target branch to throttle instead of master? Otherwise commits get tangled up when trying to merge upwards

PR #578, opened 2021-02-19
Added feature to register ECDSA hash algs separately from curves

abkarcher:
Also, updated version


app/app_main.c
```diff
@@ -536,7 +536,7 @@ static int enable_aes(ACVP_CTX *ctx) {
     CHECK_ENABLE_CAP_RV(rv);
     rv = acvp_cap_sym_cipher_set_parm(ctx, ACVP_AES_CBC_CS1, ACVP_SYM_CIPH_KEYLEN, 256);
     CHECK_ENABLE_CAP_RV(rv);
-    rv = acvp_cap_sym_cipher_set_domain(ctx, ACVP_AES_CBC_CS1, ACVP_SYM_CIPH_PTLEN, 128, 65536, 256);
+    rv = acvp_cap_sym_cipher_set_domain(ctx, ACVP_AES_CBC_CS1, ACVP_SYM_CIPH_PTLEN, 256, 65536, 64);
```
bfussell, line 539: Did these need changing  ?

bfussell (approved):

app/app_main.c
abkarcher: Yes, the CSX algs need payload lens that are not evenly divisible by AES block size according to server, and theres some implication in general that (max - min) % increment should equal 0

PR #580, opened 2021-02-23
Re-architect enums, fix some warnings

bfussell:
Create sub-cipher enums to help address cflag -Wswitch-enum
Create DOMAIN enums for similar reasons
Fix some other warnings that were exposed when using --enable-cflags configure option
Modified UT accordingly

Feel free to critique the re-architecture, I'm open for suggestions. Converting enums to enums was trickier than I imagined.

### DIFF
--- a/src/acvp_capabilities.c
+++ b/src/acvp_capabilities.c
@@ -993,6 +993,11 @@ static ACVP_RESULT acvp_validate_sym_cipher_parm_value(ACVP_CIPHER cipher, ACVP_
         case ACVP_KAS_FFC_NOCOMP:
         case ACVP_KAS_KDF_ONESTEP:
         case ACVP_KAS_HKDF:
+        case ACVP_KAS_FFC_SSC:
+        case ACVP_KAS_IFC_SSC:
+        case ACVP_KTS_IFC:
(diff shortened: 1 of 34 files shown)


src/acvp_capabilities.c
```diff
@@ -993,6 +993,11 @@ static ACVP_RESULT acvp_validate_sym_cipher_parm_value(ACVP_CIPHER cipher, ACVP_
         case ACVP_KAS_FFC_NOCOMP:
         case ACVP_KAS_KDF_ONESTEP:
         case ACVP_KAS_HKDF:
+        case ACVP_KAS_FFC_SSC:
```
abkarcher, line 996: Do we still use the fulll ACVP_CIPHER enum in these places because TDES and AES are different sub-enums? Should we make them the same sym_ciph sub enum, or move towards making tdes and aes separate structures?

bfussell: Exactly, we could dig into that later and it may save a bunch of space on the sym APIs however it would also add more entries to AES and TDES. I didn't dig into much but it may end up being about equal.

abkarcher (approved):

PR #621, opened 2021-05-24
Fixed function call in test cases

abkarcher:
Not quite sure how I didn't get this error on my platform.


bfussell (approved):
Happens, np :)

PR #630, opened 2021-07-01
BIg documentation updates, remove prebuilt doc, lots more acvp.h corrections

abkarcher:
- Deleted prebuilt doc
- Added "make doc" command which just tells doxygen (if installed) to build the documentation html and place it in the srcdir/doc folder. This is hopefully a stopgap and is likely to be polished later on if I can get automake to behave (e.g. work with make install).
- Updated Doxyfile to a new version. Made various changes to make the generated HTML appear n a better format
- More updates to acvp.h, fixing lots of typos, references to non-existent functions, etc. Added a "group" block for all of the public APIs which are now neatly listed in their own category in the generated HTML. More updates are on the way; just splitting things up to keep it manageable.
- The html generated by doxygen is now useable; but missing much information. More updates are also on the way (some stubborn doxygen bugs to polish out).
- Made updates to readme to make doxygen content with it, as it is used as the front page


bfussell (approved):

abkarcher:
Added a new commit 1bd9cea with some more fixes. Most were already internal. These will be the last changes before the libacvp 1.4.0 release unless I find some issue in testing.

bfussell (approved):

PR #643, opened 2021-07-23
Add SSL/FOM info to version command, save request URLs to testSession file

abkarcher:
When running version command in no-runtime mode, the output will now look similar to this:
```
ACVP library version(protocol version): libacvp_oss-1.4.0(1.0)

        Runtime mode: no
 FIPS module version: FIPS 2.0.16 validated module 24 Apr 2017
Compiled SSL version: OpenSSL 1.0.2u-fips  20 Dec 2019
  Linked SSL version: OpenSSL 1.0.2u-fips  20 Dec 2019
```
When running in runtime mode, it will look similar to this. We dont set FIPS mode in runtime mode, but I feel it is worth communicating that fact to people.
```
ACVP library version(protocol version): libacvp_oss-1.4.0(1.0)

        Runtime mode: yes
           FIPS mode: no
Compiled SSL version: OpenSSL 1.0.2u-fips  20 Dec 2019
  Linked SSL version: OpenSSL 1.0.2u-fips  20 Dec 2019
```

I did not change the acvp version string itself in case ours or other people's pipelines are looking for it. 

Finally, for test sessions run with --fips_validation, the request URL that libacvp prints at the very end will also be saved to the testSession file as "validationRequestUrl". This is to help avoid issues where the request URLs get misplaced as they do not contain much context for their contents until they are approved.

### DIFF
--- a/app/app_cli.c
+++ b/app/app_cli.c
@@ -15,8 +15,10 @@
 #include "safe_lib.h"
 
 #ifdef ACVP_NO_RUNTIME
-# include "app_fips_lcl.h"
+#include "app_fips_lcl.h"
 #endif
+
+#include <openssl/crypto.h>
 
 #define ANSI_COLOR_RED "\x1b[31m"
 #define ANSI_COLOR_YELLOW "\x1b[33m"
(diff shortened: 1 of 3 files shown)


bfussell (approved):

xmotrcity (approved):
I like the change

PR #654, opened 2021-09-16
Updated to revision 2.0 for SHA3 and AES-XTS, misc. improvements

abkarcher:
- Updated SHA3 to revision 2.0. SHA3 now registers supported message lengths. Large Data testing not added yet; will make a note to look into.
- Updated AES-XTS to revision 2.0. AES-XTS can now optionally register dataUnitLengths lower than the total payload size if the implementation supports it (openssl does not)
- Various updates to syntax, define names, etc.
- Update to AES-CBC-CSX registrations to give them payload lengths that are not multiples of the AES block size (required).


app/app_main.c
```diff
@@ -1051,27 +1061,31 @@ static int enable_hash(ACVP_CTX *ctx) {
     CHECK_ENABLE_CAP_RV(rv);
     rv = acvp_cap_hash_set_parm(ctx, ACVP_HASH_SHA3_224, ACVP_HASH_IN_EMPTY, 1);
     CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_cap_hash_set_domain(ctx, ACVP_HASH_SHA3_224, ACVP_HASH_MESSAGE_LEN, 0, 65536, 8);
```
bfussell, line 1064: Do these new calls need a CHECK_ENABLE_CAP_RV(rv) afterwards ? Or is it just the isn't showing it ?

abkarcher: You're right, thanks!

bfussell (approved):

PR #658, opened 2021-10-01
Add support for TLS 1.2 KDF

abkarcher:
Due to the nature of the name/mode/revision change as well as the testing change in the new TLS 1.2 KDF revision, I opted to make it into a new cipher that largely replicates the old 1.2 KDF testing. Added warnings to the old cipher (which has been left in for now) that it will be no longer supported come 2022.

Some changes were made to the library side to make it more closely resemble TLS 1.3 KDF for the sake of consistency.

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -1294,9 +1294,21 @@ static int enable_kdf(ACVP_CTX *ctx) {
     ACVP_RESULT rv = ACVP_SUCCESS;
     int i, flags = 0;
 
-    /*
-     * Enable KDF-135
-     */
+
+    rv = acvp_cap_kdf_tls12_enable(ctx, &app_kdf_tls12_handler);
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_cap_set_prereq(ctx, ACVP_KDF_TLS12, ACVP_PREREQ_SHA, value);
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_cap_set_prereq(ctx, ACVP_KDF_TLS12, ACVP_PREREQ_HMAC, value);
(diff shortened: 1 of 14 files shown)


app/app_main.c
```diff
@@ -1294,9 +1294,21 @@ static int enable_kdf(ACVP_CTX *ctx) {
...
+    rv = acvp_cap_set_prereq(ctx, ACVP_KDF_TLS12, ACVP_PREREQ_HMAC, value);
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_cap_kdf_tls12_set_parm(ctx, ACVP_KDF_TLS12_HASH_ALG, ACVP_SHA256);
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_cap_kdf_tls12_set_parm(ctx, ACVP_KDF_TLS12_HASH_ALG, ACVP_SHA384);
+    CHECK_ENABLE_CAP_RV(rv);
+    rv = acvp_cap_kdf_tls12_set_parm(ctx, ACVP_KDF_TLS12_HASH_ALG, ACVP_SHA512);
+    CHECK_ENABLE_CAP_RV(rv);
+
+#if 0 //replaced by KDF TLS12. Testing support will be removed from NIST server at the beginning of 2022.
```
bfussell, line 1311: Is the new kdf already available on the production server ?

abkarcher: Yes! according to the email I received about it

abkarcher: from /acvp/v1/algorithms endpoint on prod:

            {
                "id": 176,
                "name": "TLS-v1.2",
                "mode": "KDF",
                "revision": "RFC7627"
            }

bfussell (approved):

PR #666, opened 2021-12-15
Renamed KAS-KDF to KDA

abkarcher:
Largely this work was done by bulk replacing all occurrences of KAS_KDF, KAS-KDF, kas-kdf, etc etc. but I have sanity looked through everything, it has all been tested to work just using the current registration string.


bfussell (approved):
Yikes ! That was a lot of find/replace. Thanks.

PR #674, opened 2022-03-11
Add support for not using a hash algorithm for Z in KAS SSC algorithms

abkarcher:
Also some misc. other changes/fixes (e.g. no longer including un-needed values for VAL test submissions, spacing)


bfussell (approved):
Good stuff ! A lot more involved than I had imagined to NOT do a hash :)

PR #687, opened 2022-03-31
TDES support, misc. mac updates, configure update

abkarcher:
Tdes now uses new IV access API. Copies to a buffer instead of providing direct pointer access to CTX.

Updated HMAC/CMAC to stop strict cflags complaining about loss of const.

Updated configure so SSL version check now comes at beginning of library detection phase instead of end. This way we dont try to detect static libraries for SSL 3.0 when enable-offline is used.

Added error check to totp callback, so libacvp doesn't just hang if app returns non-success exit code to library.

### DIFF
--- a/app/app_hmac.c
+++ b/app/app_hmac.c
@@ -15,6 +15,7 @@
 #if defined ACVP_NO_RUNTIME && OPENSSL_VERSION_NUMBER < 0x30000000L
 #include "app_fips_lcl.h"
 #endif
+#include "safe_lib.h"
 
 int app_hmac_handler(ACVP_TEST_CASE *test_case) {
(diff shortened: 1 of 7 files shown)


app/app_hmac.c
```diff
@@ -99,7 +101,9 @@ int app_hmac_handler(ACVP_TEST_CASE *test_case) {
         goto end;
     }
 
-    params[0] = OSSL_PARAM_construct_utf8_string("digest", (char*)md_name, 0);
+    mdname = calloc(256, sizeof(char)); //avoid const removal warnings
```
ashman-p, line 104: Is it worth throwing in a null check here?

ashman-p (approved):

PR #699, opened 2022-05-18
Cleanup acvp_capabilities list code (1.5.0)

abkarcher:
As I was adding twostep code (waiting to get some info back), it was clear that code related to iterating through PARAM_LISTs and NAME_LISTs was getting pretty messy. This was being done on a case-by-case basis in acvp_capabilities.c. This adds iterator functions to acvp_utils, and cleans up all those calls in acvp_capabilities, saving us a few hundred lines of code.


bfussell (approved):
That was a good idea, it had gotten out of hand.  Thanks !

PR #717, opened 2022-08-16
DRBG support on 3.0

abkarcher:
Works with most of the registration of OpenSSL 3.0. Todo: library adjustments to be able to test counter DRBG with derivation function and without in the same test session. Putting that on the backburner for time being.


bfussell (approved):
Great !

malbert1 (approved):

PR #723, opened 2022-09-23
Updated support for AES algorithms on 3.X

abkarcher:
Support added for ciphertext stealing algorithms and keywrap with padding. 
Updated API calls for several AES algorithms on 3.0 (left AEAD ciphers alone for time being)
Updated GMAC to use new API. 

Removed default printing of expected results on failure of sample test sessions; we can use --get_expected_results feature now.

Note: Most of the code was copied from 1.1.1 processing code, but updated where needed; I will go back and update with more modern API calls at a later time.


app/app_aes.c
```diff
@@ -23,6 +27,642 @@ void app_aes_cleanup(void) {
...
+    case ACVP_SUB_AES_GMAC:
+    default:
+        printf("Error: Unsupported AES mode requested by ACVP server\n");
+        rv = 1;
+        goto err;
+    }
+
+    memzero_s(alg_name, 32);
+    strncat_s(alg_name, 32, type, strnlen_s(type, 16));
+    strncat_s(alg_name, 32, mode, strnlen_s(mode, 16));
```
malbert1, line 141: does mode overwrite type ?

abkarcher: strncat works by looking for a null terminator before the end of a given buffer and concatenating starting there; so the memzero prevents that.

app/app_aes.c
```diff
@@ -23,6 +27,642 @@ void app_aes_cleanup(void) {
...
+    /* Begin encrypt code section */
+    cipher_ctx = glb_cipher_ctx;
+    if ((tc->test_type != ACVP_SYM_TEST_TYPE_MCT)) {
+        EVP_CIPHER_CTX_init(cipher_ctx);
+    }
+
+    alg = acvp_get_aes_alg(tc->cipher);
+    if (alg == 0) {
+        printf("Invalid cipher value");
+        return 1;
```
malbert1, line 67: should this be 'goto err' to free glb_cipher_ctx ?

app/app_aes.c
```diff
@@ -23,6 +27,642 @@ void app_aes_cleanup(void) {
...
+    const char *type = NULL, *mode = NULL, *inv = NULL;
+
+    if (!test_case) {
+        return rv;
+    }
+
+    tc = test_case->tc.symmetric;
+
+    /* Begin encrypt code section */
+    cipher_ctx = EVP_CIPHER_CTX_new();
```
malbert1, line 284: check for null ?

app/app_aes.c
```diff
@@ -23,6 +27,642 @@ void app_aes_cleanup(void) {
...
+    EVP_CIPHER_CTX_init(cipher_ctx);
+
+    alg = acvp_get_aes_alg(tc->cipher);
+    if (alg == 0) {
+        printf("Invalid cipher value");
+        return 1;
```
malbert1, line 290: goto err ?

abkarcher: These were all some of the bits I just copied over. Fixed in new and old, thanks!

malbert1 (approved):

PR #727, opened 2022-10-05
KTS-IFC support for 3.X, enhanced library support

abkarcher:


src/acvp_kts_ifc.c
```diff
@@ -81,23 +72,29 @@ static ACVP_RESULT acvp_kts_ifc_init_tc(ACVP_CTX *ctx,
...
     stc->test_type = test_type;
     stc->md = hash_alg;
     stc->kts_role = role;
     stc->key_gen = key_gen;
 
+    stc->ct = calloc(1, ACVP_KTS_IFC_BYTE_MAX);
+    if (!stc->ct) { return ACVP_MALLOC_FAIL; }
+
+    stc->pt = calloc(1, ACVP_KTS_IFC_BYTE_MAX);
+    if (!stc->pt) { return ACVP_MALLOC_FAIL; }
```
malbert1, line 94: if you return here, do you need to free stc->ct ?

abkarcher: the test case gets passed back to the caller which eventually frees any non-null pointers in its own function (acvp_kts_ifc_release_tc)

malbert1 (approved):

bfussell (approved):

PR #731, opened 2022-10-11
ECDSA component test support, more registration updates

abkarcher:
More registration updates to match 3.0 cert, small fix for DRBG registration building, and support for ECDSA siggen/sigver component test.

Note that component test support requires an entire other vector set, instead of supporting both types of test in one vector set like usual.

### DIFF
--- a/app/app_ecdsa.c
+++ b/app/app_ecdsa.c
@@ -58,7 +58,7 @@ int app_ecdsa_handler(ACVP_TEST_CASE *test_case) {
     ACVP_SUB_ECDSA alg;
     ACVP_ECDSA_TC *tc = NULL;
     EVP_MD_CTX *sig_ctx = NULL;
-    EVP_PKEY_CTX *pkey_ctx = NULL;
+    EVP_PKEY_CTX *pkey_ctx = NULL, *comp_ctx = NULL;
     EVP_PKEY *pkey = NULL;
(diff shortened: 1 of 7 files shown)


app/app_ecdsa.c
```diff
@@ -296,18 +314,34 @@ int app_ecdsa_handler(ACVP_TEST_CASE *test_case) {
...
+
+            if (EVP_DigestVerifyInit_ex(sig_ctx, NULL, md, NULL, NULL, pkey, NULL) != 1) {
+                printf("Error initializing signing for ECDSA sigver\n");
+                goto err;
+            }
+            if (EVP_DigestVerify(sig_ctx, sig, sig_len, tc->message, tc->msg_len) == 1) {
+                tc->ver_disposition = 1;
+            }
+        } else {
+            comp_ctx = EVP_PKEY_CTX_new_from_pkey(NULL, pkey, NULL);
```
ashman-p, line 333: Did comp_ctx just leaked from the previous assignment?

abkarcher: the previous assignment is in a difference switch case for siggen instead of sigver and breaks; so we dont hit both branches in the same test case

bfussell (approved):

malbert1 (approved):

ashman-p (approved):

PR #753, opened 2022-12-23
Fix empty inputs on entropyInput for ctrdrbg

samuel40791765:
There were some ACVP test vectors we were trying to parse, which had empty inputs for `entropyInput` in the `otherInput` vectors. The field exists, but the string length is `0`. 
I assume `!tc->pr1_len || !tc->pr2_len` are helping to check if the field exists, but because the length is `0`, this causes DRBG ACVP parsing to fail prematurely.

I found an example of such case here: https://pages.nist.gov/ACVP/draft-vassilev-acvp-drbg.html#section-13-7. Although the example is for `ACVP_DRBG_SHA_256`, such cases do exists for `ACVP_DRBG_AES_128/192/256` as well.

Tested with OpenSSL 1.0.2.

### DIFF
--- a/app/app_drbg.c
+++ b/app/app_drbg.c
@@ -229,8 +229,8 @@ int app_drbg_handler(ACVP_TEST_CASE *test_case) {
         printf("Missing entropy input needed for reseed\n");
         return 1;
     }
-    if (!drbg_entropy_len || !tc->pr1_len || !tc->pr2_len ||
-        !tc->entropy || !tc->entropy_input_pr_1 || !tc->entropy_input_pr_2) {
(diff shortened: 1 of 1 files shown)


abkarcher:
Thanks! I will take a closer look at this after the new year when I am back in the office.

Andrew

abkarcher:
Hello! 

This makes sense. Could you open this PR for the 1_5_0-throttle branch?

Thanks!
Andrew

abkarcher (approved):

samuel40791765:
Awesome, I'll do that. Thanks for looking at this! 👍

abkarcher:
Closing this one; merged into 1.5.0 throttle and will be merged into main at time of next release. Thanks again!

Andrew

PR #768, opened 2023-04-06
Updates to MS build system for 2.0

abkarcher:
Windows build system had several redundant or unnecessary build options:
SafeC build options removed (they don't apply, always using safeC stub)
Legacy SSL (<= 1.0.2) options removed
static linking options removed (Just support DLLs)

Left in offline mode builds for now.

Fully tested Running OpenSSL 3.0.X tests on windows - success! IDed an issue in KAS-IFC where a test case could output as Passed when It was supposed to fail because diffs from memcmp added up to 0 (this could have happened on any platform but since windows only returns -1/0/1 for memcmp, it made the issue manifest much more quickly and obviously. Took some time to track down!)

Note: changes to MS proj files are typically software generated; any formatting issues or anything there typically aren't from me.

### DIFF
--- a/ms/make_app.bat
+++ b/ms/make_app.bat
@@ -2,15 +2,14 @@
 
 set ACV_INC_PATHS=
 set ACV_LIB_PATHS=
-set ACV_ROOT_PATH=
 
 rem Visual Studio wants absolute paths in some cases
 set ACV_ROOT_PATH_REL=%~dp0..\
 for %%i in ("%ACV_ROOT_PATH_REL%") do SET "ACV_ROOT_PATH=%%~fi
-if [%FOM_DIR%] == [] (
-  echo "No fom, some algorithms will not be available for testing"
-  set PROJ_CONFIG=nofom
-) else (
+
+set PROJ_CONFIG="nofom"
+
+if NOT [%FOM_DIR%] == [] (
   set ACV_LIB_PATHS=%FOM_DIR%\lib
   set ACV_INC_PATHS=%FOM_DIR%\include
   set PROJ_CONFIG=fom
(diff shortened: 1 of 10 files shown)


ms/make_lib.bat
```diff
@@ -6,41 +6,24 @@ set ACV_LIB_PATHS=
...
-  set ACV_INC_PATHS=%ACV_INC_PATHS%;%SAFEC_DIR%\include
+  set ACV_LIB_PATHS=%LIBCURL_DIR%\lib
+  set ACV_INC_PATHS=%LIBCURL_DIR%\include
 )
 
 if [%LIBCURL_DIR%] == [] (
-  set PROJ_CONFIG=%PROJ_CONFIG%_murl
-  set ACV_INC_PATHS=%ACV_INC_PATHS%;%ACV_ROOT_PATH%\murl
+  if [%SSL_DIR%] == [] (
+    echo "No SSL dir specified. Must be provided for online builds. exiting..."
```
ashman-p, line 21: Remove ^M (carriage return)

ms/make_app.bat
```diff
@@ -2,15 +2,14 @@
...
 
 rem Visual Studio wants absolute paths in some cases
 set ACV_ROOT_PATH_REL=%~dp0..\
 for %%i in ("%ACV_ROOT_PATH_REL%") do SET "ACV_ROOT_PATH=%%~fi
-if [%FOM_DIR%] == [] (
-  echo "No fom, some algorithms will not be available for testing"
-  set PROJ_CONFIG=nofom
-) else (
+
+set PROJ_CONFIG="nofom"
```
ashman-p, line 10: Remove ^M

ms/make_lib.bat
abkarcher: I am a bit vague on how Git displays line endings, but .bat files need CRLF line endings for labels to work properly; unless that is not the issue here?

malbert1 (approved):

ms/make_lib.bat
ashman-p: I think they show up depending on the editor used (typically a none plan text format).
Not a big deal.

ashman-p (approved):

PR #784, opened 2023-06-07
Various bugfixes and capability fixes

RandallSteck:
Fix DRBG ptr ref error
Fix AES-GCM min IV length (8->0)
Fix arguments and enum for KTS IFC set scheme function
Set ciphers in KDA TC structs
Add blank line before EOF (some editors complain without it)

### DIFF
--- a/include/acvp/acvp_lcl.h
+++ b/include/acvp/acvp_lcl.h
@@ -77,6 +77,7 @@
 #define ACVP_REV_STR_SP800_56BR2 "Sp800-56Br2"
 #define ACVP_REV_STR_SP800_56CR1 "Sp800-56Cr1"
 #define ACVP_REV_STR_SP800_56CR2 "Sp800-56Cr2"
+#define ACVP_REV_STR_SP800_108R1 "Sp800-108r1"
 #define ACVP_REV_STR_RFC8446 "RFC8446"
 #define ACVP_REV_STR_RFC7627 "RFC7627"
 
@@ -363,6 +364,7 @@
 #define ACVP_MODE_COUNTER           "counter"
 #define ACVP_MODE_FEEDBACK          "feedback"
 #define ACVP_MODE_DPI               "double pipeline iteration"
+#define ACVP_MODE_KMAC              "KMAC"
 #define ACVP_KDF135_ALG_STR         "kdf-components"
 
 #define ACVP_AUTH_METHOD_DSA_STR "dsa"
@@ -686,6 +688,14 @@
 #define ACVP_KDF108_FIXED_DATA_BIT_MAX 512 /**< Arbitrary */
 #define ACVP_KDF108_FIXED_DATA_BYTE_MAX (ACVP_KDF108_FIXED_DATA_BIT_MAX >> 3)
 #define ACVP_KDF108_FIXED_DATA_STR_MAX (ACVP_KDF108_FIXED_DATA_BIT_MAX >> 2)
+
+#define ACVP_KDF108_CONTEXT_BIT_MAX 4096 /**< Based on supportedLengths */
+#define ACVP_KDF108_CONTEXT_BYTE_MAX (ACVP_KDF108_CONTEXT_BIT_MAX >> 3)
+#define ACVP_KDF108_CONTEXT_STR_MAX (ACVP_KDF108_CONTEXT_BIT_MAX >> 2)
+
+#define ACVP_KDF108_LABEL_BIT_MAX 4096 /**< Based on supportedLengths */
+#define ACVP_KDF108_LABEL_BYTE_MAX (ACVP_KDF108_LABEL_BIT_MAX >> 3)
+#define ACVP_KDF108_LABEL_STR_MAX (ACVP_KDF108_LABEL_BIT_MAX >> 2)
 /*
  * END KDF108
  */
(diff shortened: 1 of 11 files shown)


RandallSteck:
Second commit adds changes to acvp.h I missed in the first commit

RandallSteck:
Hash LDTs do not have test written, and I'm unlikely to complete them anytime soon.  Accept or reject, I wanted to make the commit.

I have verified these against the demo server with 3.0.8 and 3.1.

RandallSteck:
KDF108-KMAC: Ditto on test cases and app.  

I have verified these against the demo server with 3.1.

src/acvp_build_register.c
```diff
@@ -1666,6 +1679,83 @@ static ACVP_RESULT acvp_build_kdf108_register_cap(JSON_Object *cap_obj, ACVP_CAP
...
+    }
+
+    /* key derivation key length list */
+    json_object_set_value(cap_obj, "keyDerivationKeyLength", json_value_init_array());
+    tmp_arr = json_object_get_array(cap_obj, "keyDerivationKeyLength");
+    tmp_val = json_value_init_object();
+    tmp_obj = json_value_get_object(tmp_val);
+    json_object_set_number(tmp_obj, "min", params->derivation_keylens.min);
+    json_object_set_number(tmp_obj, "max", params->derivation_keylens.max);
+    json_object_set_number(tmp_obj, "increment", params->derivation_keylens.increment);
```
abkarcher, line 1723: We also want to append all the "values" in the list from ACVP_JSON_DOMAIN_OBJ to the array to stay in line with the domain definition in the spec (this hasn't been retroactively added to older algorithms yet, so it isn't very clear, its a very new change). Thanks!

include/acvp/acvp_lcl.h
```diff
@@ -1143,12 +1154,17 @@ typedef struct acvp_kdf108_mode_params {
     ACVP_SL_LIST *counter_lens;
     int empty_iv_support;
     int requires_empty_iv;
+    ACVP_JSON_DOMAIN_OBJ derivation_keylens;
```
abkarcher, line 1157: Domain objects now need (its a newer change) to be freed explicitly in acvp.c using acvp_cap_free_domain. Thanks!

RandallSteck: I posted a fix for this with the scope of the kdf108 algorithm.  I don't believe the hash LDTs used any domain-type parameters.

src/acvp_build_register.c
RandallSteck: I can't find reference to the change in the ACVP spec, nor can I find an example in code (all the *_domain_set() functions don't ref the "values" field in the domain object).  Can you help find a ref to understand the requested change?

RandallSteck:
For a little context/background, I don't use your app, just the library.  I have my own front end that I can link to any library (with appropriate code).  My app code looks different, but the capabilities registration code is basically identical.  I will try to add at least a stub of the capabilities registration code to acvp.c in the future for algorithms that I add.

As noted elsewhere, I'm working on EdDSA for quick turn-around.  After that, I'd like to add FFE,  I had to use proxy/parser for a client that needed FFE, but I'd prefer to keep everything together here.  Since FFE isn't yet in OSSL3, you won't be able to test this.  

I also have code (from a 1.something branch) to read/write ACVPProxy-type files, giving interop with the other major OSS project for ACVP.  I'll try to bring that forward in a future patch as well.

I'll post the else-changes with other patches as I move forward.

How do you feel about foundational changes to the library?  Like removing the requirement to base execution from a list of vsIds?  It appears to me to be completely arbitrary to walk the list of vsId entries (looking up the algorithm within the algorithm array object) rather than simply walk the array of algorithms.  This is the source of the majority of interop/compatibility problems with ACVPProxy.

abkarcher:
Thanks for the context!

At this point I am not looking to make foundational changes as we are launching 2.0 pretty soon. I do understand the reasoning for wanting to change it (also worth noting that libacvp was designed alongside the protocol even in the early days of acvp, so some design may seem odder now). I do have some work in progress to process "raw" vector sets (a file with a single vector set brought in from elsewhere with a basic acvVersion header) which may help interop. Not sure if that will make the 2.0 initial release.

For future additions, it may be helpful to break up multiple features into multiple PRs so we can look through them faster; planning to do some testing with this one tomorrow.

Thanks again,
Andrew

RandallSteck:
I apologize for the chaotic submissions.  I've been accumulating some of these changes for a while, and I wanted to get them to you before 3.0.0.  In some cases I'm pushing changes both left and right and few changes ended up coming over in several commits.  Now that you've had a chance to give me some basic feedback, and I've had a chance to touch up some stuff (domain free and else), I could trash this set of commits and re-send them in a more organized set if it would help.  I'm flexible.

abkarcher:
I can see the differentiation between the other SP800-108 modes and KMAC gets a little messy in the protocol.  I think the way you have it implemented is the best it can be done for now, but I have opened an issue on the ACVP github to maybe get some clarification.

The other thing I saw is that SHA3 also seems to support LDT (though shake doesn't), is there a reason to have it disabled in the library? Otherwise, ran some LDT tests myself and it was perfect.

Finally, the length list for SHA capability object needs to be freed (needs acvp_cap_free_sl call at acvp.c:672).

Rest looks great, thanks a ton!

Andrew

abkarcher (approved):

abkarcher:
Hey Randy,

We are going to go ahead and merge this (I will add a thing or two of my own in a future PR) so we can lock down some code for 2.0 release testing.

Thanks,
Andrew

PR #806, opened 2024-01-24
Use prehash instance if necessary in EDDSA sample app

0140454:
Use correct instance for EDDSA operations based on the return value from `get_ed_instance_param()`.

Although `ACVP_EDDSA_SUPPORTS_PREHASH` is false in app_main.c, I thought sample app still can try to use proper instance based on test case.


abkarcher:
Thanks! I will run a test or two and report back. I should enable prehash in app_main soon anyway.

abkarcher (approved):

bfussell (approved):

PR #810, opened 2024-02-01
FIPS 186-5 support for RSA, misc. RSA updates

abkarcher:
This one got a bit messy. The RSA capability registration code is pretty old so I brought it up to speed a little bit. It sill isn't perfect but its a little more consistent; I will make a second pass at it at some point.

- FIPS 186-5 support for RSA. 186-4 support is also maintained.
- Removed storing const string references in the RSA capability object (Just use enums and only pull strings when we are building the registration)
- New API's needed to set mask function for RSA signatures, as the other APIs didnt quite have the right combination of parameters

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -2739,6 +2739,7 @@ static int enable_rsa(ACVP_CTX *ctx) {
     CHECK_ENABLE_CAP_RV(rv);
 #ifndef ACVP_FIPS186_5
     rv = acvp_cap_rsa_siggen_set_parm(ctx, ACVP_RSA_PARM_REVISION, ACVP_REVISION_FIPS186_4);
+    CHECK_ENABLE_CAP_RV(rv);
     // RSA w/ sigType: X9.31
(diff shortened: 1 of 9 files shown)


app/app_main.c
```diff
@@ -2703,6 +2737,8 @@ static int enable_rsa(ACVP_CTX *ctx) {
     CHECK_ENABLE_CAP_RV(rv);
     rv = acvp_cap_set_prereq(ctx, ACVP_RSA_SIGGEN, ACVP_PREREQ_DRBG, value);
     CHECK_ENABLE_CAP_RV(rv);
+#ifndef ACVP_FIPS186_5
+    rv = acvp_cap_rsa_siggen_set_parm(ctx, ACVP_RSA_PARM_REVISION, ACVP_REVISION_FIPS186_4);
```
malbert1, line 2741: do you need a CHECK after this ?

malbert1 (approved):

ashman-p (approved):

bfussell (approved):

PR #813, opened 2024-02-13
Registration updates

abkarcher:
- Disable EDDSA for non FIPS186-5 modules
- AES-CBC-CS3 PT len lower bound from 136 to 128 for 3.1.2
- give AES GMAC an IV range instead of a single IV for 3.1.2
- Large data testing for SHA:
    - Enabled for SHA3 for 3.0.8 or greater by default
    - Enabled for rest of SHA for 3.1.2 or greater by default
    - Removed the configure option to enable LDT since its now needed by default
- Add more HMAC modes for KDF sp800-108 for 3.1.2 or greater
- PBKDF: add SHA3 support for 3.1.2 or greater
- KAS-FFC: Add MODP modes back for 3.1.2 or greater
- Safe Primes: Add MODP modes back for 3.1.2 or greater
- ECDSA sigs: add SHA3 for 3.1.2 or greater
- RSA bugfix

More registration updates incoming, just the first batch

### DIFF
--- a/app/app_main.c
+++ b/app/app_main.c
@@ -146,6 +146,11 @@ static int verify_algorithms(APP_CONFIG *cfg) {
         CHECK_NON_ALLOWED_ALG(cfg->safe_primes, "This version of OpenSSL does not support safe primes testing");
     }
 #endif
+#ifndef ACVP_FIPS186_5
+    if (!cfg->testall) {
+        CHECK_NON_ALLOWED_ALG(cfg->eddsa, "This version of OpenSSL does not support DSA testing (DSA disabled)");
+    }
+#endif
 
     return rv;
 }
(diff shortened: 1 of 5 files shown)


bfussell (approved):

app/app_main.c
```diff
@@ -2142,7 +2182,7 @@ static int enable_kas_ffc(ACVP_CTX *ctx) {
     CHECK_ENABLE_CAP_RV(rv);
     rv = acvp_cap_kas_ffc_set_parm(ctx, ACVP_KAS_FFC_SSC, ACVP_KAS_FFC_MODE_NONE, ACVP_KAS_FFC_GEN_METH, ACVP_KAS_FFC_FB);
     CHECK_ENABLE_CAP_RV(rv);
-#if OPENSSL_VERSION_NUMBER < 0x30000000L
+#if OPENSSL_VERSION_NUMBER >= 0x30100000L
```
malbert1, line 2185: originally you wanted an old version, and now you want the newest version ?

abkarcher: Correct. Previously, we wanted to only use with FOM and 1.X.X then skip 3.0 since it wasnt on the 3.0 cert, but now its on the 3.1.2 cert :) some fragments of 1.X.X support remain but I am slowly stripping them out.

malbert1 (approved):

PR #814, opened 2024-02-14
More registration updates for 3.1.X

abkarcher:
- truncated SHA for PKCS 1.5 rsa sigs
- table C3 added for rsa keygen
- KDA onestep: added ALL aux functions to 3.1.X, kept the small subset for 3.0.X, removed most 1.X.X support
- GMAC registration fix
- KDA onestep aux function fixes

Should have one more PR after this to add KMAC KDF


app/app_main.c
ashman-p, line 1: Lots if #ifs. I t might help clarify the pairs with comments next time you are in the file.

abkarcher: Agreed, its getting pretty messy. I plan to start splitting different versions' registrations up into different source files in a future release, but if I don't, I will definitely start adding comments. Thanks!

bfussell (approved):

PR #835, opened 2024-06-28
Functional enhancements

RandallSteck:
Enhance vector processing -- allow processing of acvpproxy-style vector files and extract vsId from filename if not found in file

Add domain option to several symmetric parameters

Minor other tweaks and typos


RandallSteck:
The new "offline" vector processing function is so named because it ignores the header (jwt, vsid, etc) and gets right to processing whatever it finds.  It will process an array of algorithms (libacvp-style) or a single algorithm (acvpproxy style).  It is called "offline" since we're not tracking any of the session, it is probably not usable (and indeed isn't required) in the case of a full single-command round-trip testing session.

Add a bit of recursive file walking and you can chew through a ton of vector files for little additional work.

RandallSteck:
I forgot to include acvp_lcl.h.  I will close this and resubmit.

PR #836, opened 2024-06-29
Functional Enhancements (v2)

RandallSteck:
Enhance vector processing -- 
1) Modify main vector processor to extract vsId from filename if not found in file.  Proxy-style files do not have vsId in the file, and libacvp requires vsId for processing.  If a vsId does not exist in the file, the filename will be searched for "vs" followed by digits, which will be interpreted as a vsId.  This was an early patch for proxy-style files that works well enough.

2) New function to process vectors that specifically supports acvpproxy-style vector files.  It is a 1st attempt at a "universal" vector file processor.  The function is named "offline" because it ignores the header (jwt, vsid, etc) and gets right to processing whatever it finds. It will process an array of algorithms (libacvp-style) or a single algorithm (acvpproxy style).  It is resilient to the occasional degenerate case of a naked (non-arrayed) json object. It is called "offline" since we're not tracking any of the session data.  The function is probably not usable (and indeed isn't required) in the case of a single-command full-round-trip testing session. That said, I have not blocked it off with ACVP_OFFLINE.  

Add domain-type for PT length in symmetric ciphers -- Protocol allows domains, libacvp allowed only value arrays.

Minor other tweaks and typos

### DIFF
--- a/src/acvp.c
+++ b/src/acvp.c
@@ -1020,6 +1020,7 @@ ACVP_RESULT acvp_run_vectors_from_file(ACVP_CTX *ctx, const char *req_filename, 
     int vs_cnt = 0, isSample = 0;
     const char *jwt = NULL;
     char *json_result = NULL;
+    char url_buffer[128] = { 0 };  // for vsId from filename
 
     ACVP_LOG_STATUS("Beginning offline processing of vector sets...");
 
@@ -1056,7 +1057,9 @@ ACVP_RESULT acvp_run_vectors_from_file(ACVP_CTX *ctx, const char *req_filename, 
         strcpy_s(ctx->session_url, ACVP_ATTR_URL_MAX + 1, test_session_url);
     } else {
         ACVP_LOG_WARN("Missing session URL, results will not be POSTed to server");
+#ifndef ACVP_OFFLINE
         goto end;
+#endif
     }
 
     jwt = json_object_get_string(obj, "jwt");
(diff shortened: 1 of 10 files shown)


src/acvp.c
```diff
@@ -1089,6 +1094,20 @@ ACVP_RESULT acvp_run_vectors_from_file(ACVP_CTX *ctx, const char *req_filename,
         if (rv != ACVP_SUCCESS) goto end;
         ACVP_LOG_INFO("Received vsid_url=%s", vsid_url);
     }
+    /* If no vsIds found, try to get it from the filename */
+    if (vs_cnt == 0) {
+        int vsid;
+        char* start = strstr(req_filename, "vs");
+        if (!start || (0 == (vsid = atoi(start+2)))) {
```
abkarcher, line 1101: could we add a check to make sure start+2 does not go past end of string?

src/acvp_util.c
```diff
@@ -1563,10 +1563,12 @@ ACVP_RESULT acvp_json_serialize_to_file_pretty_a(const JSON_Value *value, const
             fclose(fp);
             return ACVP_JSON_ERR;
         }
+        if (!first) {  // Do not add "," before this data
```
abkarcher, line 1566: can we combine these nested if statements? static analysis will complain :)

abkarcher:
the "universal" vector set parsing works fantastically, thanks! After we sort out the other comments, can you re-open this PR against 2_2_0-throttle? Typically, we want to keep API additions and changes to new minor release versions when possible.

Thanks,
Andrew

RandallSteck:
Closed to be re-open against 2_2_0

PR #837, opened 2024-07-09
Functional enhancements (with req fixes)

RandallSteck:
Enhance vector processing --

Modify main vector processor to extract vsId from filename if not found in file. Proxy-style files do not have vsId in the file, and libacvp requires vsId for processing. If a vsId does not exist in the file, the filename will be searched for "vs" followed by digits, which will be interpreted as a vsId. This was an early patch for proxy-style files that works well enough.

New function to process vectors that specifically supports acvpproxy-style vector files. It is a 1st attempt at a "universal" vector file processor. The function is named "offline" because it ignores the header (jwt, vsid, etc) and gets right to processing whatever it finds. It will process an array of algorithms (libacvp-style) or a single algorithm (acvpproxy style). It is resilient to the occasional degenerate case of a naked (non-arrayed) json object. It is called "offline" since we're not tracking any of the session data. The function is probably not usable (and indeed isn't required) in the case of a single-command full-round-trip testing session. That said, I have not blocked it off with ACVP_OFFLINE.

Add domain-type for PT length in symmetric ciphers -- Protocol allows domains, libacvp allowed only value arrays.

Two fixes to address comments from [PR#836](https://github.com/cisco/libacvp/pull/836)

Other minor tweaks and typos


src/acvp.c
```diff
@@ -1097,8 +1097,9 @@ ACVP_RESULT acvp_run_vectors_from_file(ACVP_CTX *ctx, const char *req_filename,
     /* If no vsIds found, try to get it from the filename */
     if (vs_cnt == 0) {
         int vsid;
+        int len = strnlen_s(req_filename, ACVP_JSON_FILENAME_MAX);
         char* start = strstr(req_filename, "vs");
-        if (!start || (0 == (vsid = atoi(start+2)))) {
+        if (!start || (start >= len+2) || (0 == (vsid = atoi(start+2)))) {
```
abkarcher, line 1102: Something doesn't feel right here. `start` is a memory address and `len` is a small length int. maybe `strnlen_s` should be run on `start`, and then check if `len <= 2` ?

RandallSteck: Good find, thanks.  Updated and ready.

abkarcher: The acvpro_ prefix for strnlen_s does not exist in our code or compile - at the risk of being a pest :)

RandallSteck: Sorry!  Done

abkarcher (approved):

malbert1 (approved):

bfussell (approved):

ashman-p (approved):

PR #866, opened 2024-09-24
Fix ECDSA in FIPS 186-5

iomartin:
Several curves and hashes were missing for ECDSA in FIPS 186-5.

Also included two smaller changes in this PR, one to .gitignore and one that shows failed algorithms more clearly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,9 +45,11 @@
 
 # IDE files
 **/.vscode/
+*.swp
 
 # Output Directories
 **/.libs/
 
 # Misc
 *.gcov
+testSession*.json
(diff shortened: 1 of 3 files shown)


abkarcher:
Hello,

Just some extra info - The 186-5 flag was not originally intended to be compiled as much as it was to be an example for people who were needing to reference it for their own modules. The flag will be removed for the 2.2.0 release, as we have refactored the app to separate all the registrations.

Still, I am happy to accept this PR for the imminent 2.1.2 release as long as it doesn't change the capabilities registration for non 186-5 modules (e.g. existing OpenSSL FPs < 3.1.2). Could you re-open it against the 2_1_0-throttle branch?

Thanks,
Andrew

iomartin:
Thanks for reviewing and the extra info.

I have reopened this as #867 based on the  2_1_0-throttle branch

PR #903, opened 2025-06-25
First part of adding alternate MCT support to libacvp

abkarcher:
This is the first of two PRs to add support for the alternate monte-carlo testing mode for hash for libacvp.

- Adds registration support for MCT version (standard or alternate - standard done by default)
- Moves most MCT logic out of acvp_hash.c into a new source file, acvp_hash_mct.c (minor adjustments as needed)
- Adds common function for MCT test cases to delegate out to the correct MCT functions
- Tested with standard mode which continues to work with no app changes

Next PR will:
- Add actual functions and logic for running the alternate mode
- adjust any existing UTs if needed


malbert1 (approved):

abkarcher:
Much of this seems to need to be adjusted; I should have read the spec more closely before implementing. Next PR will address.

PR #910, opened 2025-07-11
Add support for multiple combinations of IV sources/modes for AES-GCM and related

abkarcher:
- Previously, supported only one IV gen mode per registration, either 8.2.1 or 8.2.2
- A module can support both
- But, a module can also support only certain IV sources (internal or external) on certain modes. so we have to track which sources can be used with which modes

This PR expands on the AES capabilities registration:
- capability object now has a "matrix" of iv mode/source combinations
- enables registration of both modes at once using the traditional param set API
- Adds a new API for specifying a specific combination of an IV source and mode, acvp_cap_sym_cipher_set_iv_modes()
- Backwards compatible with existing API calls
- Written to somewhat-easily expand if more IV modes are added to AES later on

This means, at time of writing, one AES-GCM registration can create up to 4 AES-GCM vector sets.


RandallSteck:
I can't easily swap library versions, so I won't be able to test this patch live for a few weeks.  I've reviewed the code, and it looks good based on the design you describe.  

THANK YOU for addressing this issue so quickly.  I look forward to integrating this change.

Randy

malbert1 (approved):

PR #920, opened 2025-07-29
Fix typos; Fix/update error messages

RandallSteck:
I recently synced and saw updates to some error message to remove the "ERROR: " prefix.  I've carried on that work to all files.  I've also updated various error messages for capitalization and standardization.

This is all just changes to error messages and a few comments, so it shouldn't be a hard review.


abkarcher:
Looks great, thank you!

PR #964, opened 2026-06-01
Fix ML-DSA internal MU tests

### LINKED ISSUE #963: ML-DSA sign/verify fails with internal signature and mu
Currently acvp_app only tests ML-DSA sign/verify with `ACVP_ML_DSA_PARAM_MU` set to `ACVP_ML_DSA_MU_EXTERNAL` (externalMu == True).

If this is changed to `ACVP_ML_DSA_MU_INTERNAL`, then tests fail.

Tested with openssl 3.5.6

iomartin:
This PR fixes a couple of issues with ML-DSA vectors.

The first, as described in https://github.com/cisco/libacvp/issues/963, happens when we try to use internal MU for ML-DSA. This was caused because OpenSSL was encoding the message, which it shouldn't.

The second problem was caused by the order in which test groups were executed. When reading the request JSON, setting one of the variables could be skipped, which caused it to keep the value from the previous iteration, thus producing the wrong result.

With these two fixes, we can switch mu testing from EXTERNAL to BOTH.

It was tested with:

```
acvp_app --ml_dsa
```

using OpenSSL 3.5.0.

Resolves: #963

### DIFF
--- a/app/implementations/openssl/3/iut_ml_dsa.c
+++ b/app/implementations/openssl/3/iut_ml_dsa.c
@@ -172,6 +172,11 @@ int app_ml_dsa_handler(ACVP_TEST_CASE *test_case) {
             msg_ptr = tc->mu;
             msg_len = (size_t)tc->mu_len;
             OSSL_PARAM_BLD_push_int(pbld, OSSL_SIGNATURE_PARAM_MU, 1);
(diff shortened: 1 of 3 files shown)


abkarcher:
Hello,

Thank you very much for the explanation and fix! I am busy on some other work presently but wanted to drop in and acknowledge this - I will take a look at it in the next week or so hopefully.

Thanks,
Andrew

